PR #19, opened 2014-10-04
Cleanup

tamird:
@igrigorik fixes a bunch of style violations, runs CI against JRuby and RBX, though RBX doesn't pass yet.


tamird:
@igrigorik rebased, should be good to re-review

lib/http/2/compressor.rb
```diff
@@ -134,9 +130,9 @@ def dup
       # @return [Array] +[key, value]+
       def dereference(index)
         # NOTE: index is zero-based in this module.
-        STATIC_TABLE[index] or
-          @table[index - STATIC_TABLE.size] or
-          raise CompressionError.new("Index too large")
+        STATIC_TABLE[index] ||
```
igrigorik: Are we sure about this change? @mad-p can you sanity check?

lib/http/2/compressor.rb
```diff
@@ -359,15 +349,15 @@ def set_table_size(size)
       # @return [String] binary string
       def integer(i, n)
         limit = 2**n - 1
-        return [i].pack('C') if (i < limit)
+        return [i].pack('C') if i < limit
 
         bytes = []
-        bytes.push limit if !n.zero?
+        bytes.push limit unless n.zero?
```
igrigorik: Maybe it's just me, but my brain has a _really_ hard time grokking `unless`.. I can read the `if !expr` without any issues, but I always stumble across `unless` and have to reparse it a few times before it makes sense. 

What do the various style guides recommend for unless? Personally, I'd prefer to avoid it, but maybe that's just me. /cc @mad-p

lib/http/2/compressor.rb
```diff
@@ -513,10 +503,9 @@ def string(buf)
         huffman = (buf.readbyte(0) & 0x80) == 0x80
         len = integer(buf, 7)
         str = buf.read(len)
-        str.bytesize == len or raise CompressionError.new("string too short")
-        huffman and str = Huffman.new.decode(Buffer.new(str))
-        str = str.force_encoding('utf-8')
-        str
+        str.bytesize == len || fail(CompressionError, 'string too short')
+        huffman && str = Huffman.new.decode(Buffer.new(str))
```
igrigorik: Same as before, is this a safe change? /cc @mad-p

lib/http/2/compressor.rb
tamird: The only difference between `or` and `||` is operator precedence - since there are no other operators involved in this expression, this change should be safe.

As a side note, `or`'s precedence is surprisingly low, which is why RuboCop suggests to avoid it.

lib/http/2/compressor.rb
tamird: RuboCop's style guide suggests to avoid both:
- `if !predicate` (because it's not idiomatic ruby)
- `unless predicate; .. else` (because unless-else is hard to parse)

That's the middle ground https://github.com/bbatsov/ruby-style-guide has chosen. Does that make sense?

lib/http/2/compressor.rb
tamird: For the same reasons as above, this is safe. That is, `and`/`&&` only differ in operator precedence relative to other operators, and in this case no other operators are involved.

igrigorik:
@tamird thanks for updating this.. overall, looks good! Two things that worry me: use of `unless` and `and/or` vs `&& / ||`... The former may be a personal preference (and inability of my own brain to parse `unless`), whereas the latter is more of a functional question.. since those are not interchangeable -- in theory, tests should have us covered, but I'd still like to get a +1 from @mad-p.

tamird:
@igrigorik sure, sounds good. I tried to address your concerns in line. Thanks for reviewing!

igrigorik:
@tamird thanks. I'm ok with merging this. @mad-p any thoughts/comments?

tamird:
@mad-p ping

lib/http/2/compressor.rb
mad-p: I prefer `unless` over `if negative-predicate` when it's postfix `if/unless`.
That style-guide looks good and well-accepted (though I disagree on some minor points).

lib/http/2/compressor.rb
mad-p: This particular point is what I disagree with https://github.com/bbatsov/ruby-style-guide .
I prefer `and`/`or` over `&&`/`||` when its all about flow control. `&&`/`||` should convey value, not control IMHO.
- condition `and` action
- condition `&&` condition
I prefer `or` here.  However, rewriting with nested if with condition `||` condition would be better in this case.

I tend to mix `&&` and `and` in a single sentence and it's a bad habit. I should avoid it.
- condition `||` condition `and` action
This should be fixed.

lib/http/2/compressor.rb
mad-p: These examples are all condition `and` action style.  I prefer `and`/`or`.
Probably we might want `if`/`unless` solution here:
- `raise ... unless str.bytesize == len`
- `str = Huffman.new(...) if huffman`

「condition `and` action」 vs 「action `if` condition」 is another question, but I don't want to dive into this argument here.

BTW, it's a surprise that `condition && var = value` works!  Ruby's parser seems to be too clever.
I expect it parses as `(condition && var) = value` and gives a syntax error.
`x = a && b` parses into ` x = (a && b)` but `a && b = x` becomes `(a) && ( b = x )`.
That's weird.
This behavior is Ruby-only and not consistent with Perl and some other. (AFAIK)

http-2.gemspec
```diff
@@ -4,19 +4,19 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
...
+  spec.homepage      = 'https://github.com/igrigorik/http-2'
+  spec.license       = 'MIT'
 
-  spec.files         = `git ls-files`.split($/)
-  spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
-  spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
-  spec.require_paths = ["lib"]
+  spec.files         = `git ls-files`.split($INPUT_RECORD_SEPARATOR)
+  spec.executables   = spec.files.grep(/^bin\//) { |f| File.basename(f) }
+  spec.test_files    = spec.files.grep(/^(test|spec|features)\//)
```
mad-p: `\/` looks ugly.  `%r{ ... }` syntax is there for that exact purpose.

lib/http/2/compressor.rb
tamird: This is a good point - I've converted all the `condition && do_something` cases to `do_something if condition`

http-2.gemspec
tamird: Agreed - this can be fixed on the next rubocop release which will contain: https://github.com/bbatsov/rubocop/commit/e3eca69269fd9191a2315cdbd90adae98e32340b

lib/http/2/connection.rb
```diff
@@ -464,25 +461,22 @@ def validate_settings(role, settings)
...
-        if frame[:flags].include?(:ack)
-          # Process pending settings we have sent.
-          [@pending_settings.shift, :local]
-        else
-          check = validate_settings(@remote_role, frame[:payload])
-          check and connection_error(check)
-          [frame[:payload], :remote]
-        end
+      settings, side = if frame[:flags].include?(:ack)
+        # Process pending settings we have sent.
```
mad-p: Does this change improve readability?

tamird: the `Lint/EndAlignment` cop wanted this change, I believe

lib/http/2/huffman.rb
```diff
@@ -50,283 +43,282 @@ def decode(buf)
...
-          raise CompressionError.new('Huffman decode error (EOS invalid)')
+          fail CompressionError, 'Huffman decode error (EOS invalid)'
         end
-        emit.force_encoding(BINARY)
+        emit.force_encoding(Encoding::BINARY)
       end
 
       # Huffman table as specified in
       # - http://tools.ietf.org/html/draft-ietf-httpbis-header-compression-10#appendix-B
       CODES = [
```
mad-p: This table was aligned so that easy to cut-n-paste from the spec.
https://tools.ietf.org/html/draft-ietf-httpbis-header-compression-12#appendix-B

lib/http/2/huffman_statemachine.rb
```diff
@@ -10,263 +10,278 @@ class Huffman
       # :nodoc:
       MAX_FINAL_STATE = 7
       MACHINE = [
-        [[nil,16],[nil,76],[nil,54],[nil,36],[nil,30],[nil,28],[nil,24],[nil,221],[nil,9],[nil,10],[nil,11],[nil,12],[nil,13],[nil,14],[nil,15],[nil,1],],
```
mad-p: Don't bother style for machine-generated code.

lib/http/2/stream.rb
```diff
@@ -72,12 +71,12 @@ class Stream
     # @param exclusive [Boolean]
     # @param window [Integer]
     # @param parent [Stream]
-    def initialize(connection: nil, id: nil, weight: 16, dependency: 0, exclusive: false, parent: nil)
-      @connection = connection or raise ArgumentError.new("missing mandatory argument connection")
-      @id = id                 or raise ArgumentError.new("missing mandatory argument id")
+    def initialize(connection:, id:, weight: 16, dependency: 0, exclusive: false, parent: nil)
```
mad-p: Good catch! :sushi:

lib/http/2/huffman_statemachine.rb
tamird: I managed to tweak the generation script to emit rubocop-accepted code, what's the harm?

lib/tasks/generate_huffman_table.rb
```diff
@@ -119,16 +122,39 @@ class Huffman
...
-          (1 << BITS_AT_ONCE).times do |t|
-            emit = n.transitions[t].emit
-            emit == EOS or emit = emit.dup.force_encoding('binary')
-            f.print %Q/[#{emit == '' ? "nil" : emit.inspect},#{state_id[n.transitions[t].node]}],/
-          end
+          f.print '        ['
+          string = (1 << BITS_AT_ONCE).times.map do |t|
+            transition = n.transitions.fetch(t)
+            emit = transition.emit
+            emit = emit.dup.force_encoding(Encoding::UTF_8) unless emit == EOS
```
mad-p: Why UTF-8? We are talking about binary octets rather than character string.

lib/tasks/generate_huffman_table.rb
```diff
@@ -119,16 +122,39 @@ class Huffman
...
+            when String
+              if emit.empty?
+                'nil'
+              else
+                if emit.valid_encoding?
+                  case emit.bytes
+                  when SINGLE_QUOTE, CARRIAGE_RETURN
+                    emit.inspect
+                  when SLASH
+                    "'\\\\'"
```
mad-p: Why not use `emit.inspect`? `inspect` guarantees it's parsable.
Seems like you are writing code to make the generated code less-readable-but-RoboCop-friendly.

lib/tasks/generate_huffman_table.rb
```diff
@@ -136,22 +162,22 @@ class Huffman
       end
     end
 
-    def self.root
-      @root
+    class << self
+      attr_reader :root
```
mad-p: Nice :sushi:

mad-p:
Overall, good job @tamird! Thanks a lot. 
`and/or` vs `&&/||` thing is functionally no problem (to my surprise :stuck_out_tongue: ).

About the genarated Huffman decoding table, is it possible to generate "robocop:disable"?
That might be simpler.

tamird:
@igrigorik ping?

lib/tasks/generate_huffman_table.rb
tamird: right, this is handled below in the `valid_encoding?` check

lib/tasks/generate_huffman_table.rb
igrigorik: +1

tamird: pushed a commit to address this

Gemfile
```diff
@@ -4,13 +4,15 @@ gem 'rake'
 gem 'yard'
 
 group :test do
-  gem 'coveralls', :require => false
-  gem 'rspec'
-  gem 'rspec-autotest'
-  gem 'autotest-standalone'
+  gem 'activesupport'
```
igrigorik: Is this actually used?

http-2.gemspec
igrigorik: Can we revert this then? Seems like unnecessary code churn on our end :)

Gemfile
tamird: yeah, it's needed for `deep_dup`, which was needed to make the tests work after i `freeze`d everything

http-2.gemspec
tamird: then we'll need to introduce churn to the rubocop config...

igrigorik:
@tamird thanks for your patience with this one :)

Overall, I'm OK with merging this. I may not agree with all of the style changes, but that's the nature of it.. it's style, we can bikeshed it forever. @mad-p if you're OK with all of the proposed changes, feel free to hit the merge button.

mad-p:
My last concern is regex `%r{...}` thing.
I think we can wait for the new RuboCop.

tamird:
@igrigorik @mad-p updated for latest rubocop

igrigorik:
@tamird thanks for your patience with this one, much appreciated! :)

tamird:
Thanks!

Issue #26, opened 2014-12-01
Implement draft-16


igrigorik:
Diff: http://www.ietf.org/rfcdiff?url2=draft-ietf-httpbis-http2-16

Issue #25, opened 2014-12-01
Implement HPACK-10


igrigorik:
Diff: http://www.ietf.org/rfcdiff?url2=draft-ietf-httpbis-header-compression-10

PR #28, opened 2015-01-26
Implement draft-16

mad-p:
Re: #25, #26 

draft-15 changes
* Add http_1_1_required error code

draft-16 changes
* PRIORITY can be sent/received in any state
* Update hpack-test-case to lastest (though no normative changes included)
* Examples advertise h2-16

Bugs found in added tests are fixed, too.

### DIFF
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -258,6 +258,7 @@ module HTTP2
               event(:open)
             end
           when :rst_stream then event(:local_rst)
+          when :priority then process_priority(frame)
           else stream_error; end
         else
           case frame[:type]
@@ -268,6 +269,7 @@ module HTTP2
             else
               event(:open)
             end
+          when :priority then process_priority(frame)
           else stream_error(:protocol_error); end
         end
 
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -211,48 +211,57 @@ module HTTP2
     private
 
     # HTTP 2.0 Stream States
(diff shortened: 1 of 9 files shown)


lib/http/2/stream.rb
```diff
@@ -98,14 +99,14 @@ def receive(frame)
       case frame[:type]
       when :data
         # TODO: when receiving DATA, keep track of local_window.
```
igrigorik: we can drop the TODO now, correct?

lib/http/2/stream.rb
```diff
@@ -98,14 +99,14 @@ def receive(frame)
...
+        @local_window -= frame[:payload].size
         emit(:data, frame[:payload]) if !frame[:ignore]
       when :headers, :push_promise
         emit(:headers, frame[:payload]) if !frame[:ignore]
       when :priority
         process_priority(frame)
       when :window_update
-        @remote_window += frame[:increment]
-        send_data
+        process_window_update(frame) if !frame[:ignore]
```
igrigorik: Is the separate method call necessary? Seems like we're only calling this method from just this one place? Also the if check is duplicated.

lib/http/2/stream.rb
mad-p: fixed

lib/http/2/stream.rb
mad-p: Intended to be called from connection.
Just removed duplicate :ignore check.

igrigorik:
awesome work, as usual! :+1:

PR #34, opened 2015-07-23
jruby 9k released

ferdinandrosario:
since jruby 9k out modified the travis to check http-2 is compatible with the latest ruby version.


igrigorik:
thanks!

PR #37, opened 2015-08-10
Fixes missing frame[:flags] when data frames are split

tbetbetbe:
Also fixes

a similar issue in flow_buffer#send_data

### DIFF
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -174,16 +174,16 @@ module HTTP2
     # @param end_stream [Boolean] indicates last response DATA frame
     def data(payload, end_stream: true)
       flags = []
-      flags << :end_stream if end_stream
 
       # Split data according to each frame is smaller enough
       # TODO: consider padding?
       max_size = @connection.remote_settings[:settings_max_frame_size]
       while payload.bytesize > max_size
         chunk = payload.slice!(0, max_size)
-        send(type: :data, payload: chunk)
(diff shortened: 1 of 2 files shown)


igrigorik:
When does this become an error? Do you have an error test case? Instead of doing the check should we guarantee that :flags is is always initialized?

tbetbetbe:
> When does this become an error ?
when trying to determine the transition for a DATA frame that has no flags.  E.g, when sending multiple DATA frames, any frame that is not the final one

> Do you have an error test case ?
I just added one

> Instead of doing the check should we guarantee the :flags is always initialized

That might be better
```sh
$ grep :flags -R lib | wc -l  # check for number of sites where :flags is used
   21
```
I see 21 sites where :flags is used, making sure flags is initialized would ensure that none of these are affected.  I can change the PR to do that if you like?

tbetbetbe:
The test case can be improved - maybe it should demonstrate that a parsed frame can result in a frame hash where :flags is nil, the fix would be to modify the parsing code to ensure this does not happen.

igrigorik:
> The test case can be improved - maybe it should demonstrate that a parsed frame can result in a frame hash where :flags is nil, the fix would be to modify the parsing code to ensure this does not happen.

The fact that there are 21 spots where this (flags == nil) is a potential issue is a strong argument for above approach. Let's do it :)

tbetbetbe:
Done:  PTAL

igrigorik:
:+1: great stuff!

PR #38, opened 2015-08-10
Fixes sending of rst_stream

tbetbetbe:
- rst_stream was not being sent unless there was a protocol error
- tests were broken by this change:
  * so fixed the broken tests by renaming StreamError -> InternalError
  * This was necessary because
    + stream#stream_error defaulted to calling close with :stream_error
    + :stream_error is not a DEFINED_ERROR, and could not be converted
      into a valid exception class

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -342,7 +342,12 @@ module HTTP2
         # An endpoint can end a connection at any time. In particular, an
         # endpoint MAY choose to treat a stream error as a connection error.
         if frame[:type] == :rst_stream
-          goaway(frame[:error]) if frame[:error] == :protocol_error
+          if frame[:error] == :protocol_error
+            goaway(frame[:error])
+          else
+            frames = encode(frame)
+            frames.each {|f| emit(:frame, f) }
+          end
         else
           # HEADERS and PUSH_PROMISE may generate CONTINUATION
           frames = encode(frame)
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -341,15 +341,11 @@ module HTTP2
       else
         # An endpoint can end a connection at any time. In particular, an
         # endpoint MAY choose to treat a stream error as a connection error.
-        if frame[:type] == :rst_stream
-          if frame[:error] == :protocol_error
-            goaway(frame[:error])
-          else
-            frames = encode(frame)
-            frames.each {|f| emit(:frame, f) }
-          end
+        if frame[:type] == :rst_stream && frame[:error] == :protocol_error
+          goaway(frame[:error])
         else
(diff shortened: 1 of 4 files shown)


lib/http/2/connection.rb
```diff
@@ -342,7 +342,12 @@ def send(frame)
         # An endpoint can end a connection at any time. In particular, an
         # endpoint MAY choose to treat a stream error as a connection error.
         if frame[:type] == :rst_stream
-          goaway(frame[:error]) if frame[:error] == :protocol_error
```
igrigorik: This duplicates same code as else clause. We can make this cleaner.

lib/http/2/error.rb
```diff
@@ -25,7 +25,7 @@ class FlowControlError < ProtocolError; end
 
     # Raised on invalid stream processing: invalid frame type received or
     # sent, or invalid command issued.
-    class StreamError < ProtocolError; end
+    class InternalError < ProtocolError; end
```
igrigorik: Why is this necessary? StreamError is a much better name indicating the actual type of error.

tbetbetbe: From Stream#stream_error
If stream_error is supposed to be a defined error, then we can add it an remove this change, I simply assumed it was not.

```ruby

    def stream_error(error = :stream_error, msg: nil)
      @error = error
      close(error) if @state != :closed   # Here, IUC, error needs to be a valid defined error

      klass = error.to_s.split('_').map(&:capitalize).join
      fail Error.const_get(klass), msg  # Here, the Exception class name needs to match the error name
    end
```

igrigorik: Ah, ok, I'm with you now -- lgtm.

igrigorik:
@tbetbetbe can we eliminate the duplication in https://github.com/igrigorik/http-2/pull/38#discussion_r36664031?

@mad-p any thoughts on this one?

lib/http/2/connection.rb
tbetbetbe: Fixed

igrigorik:
:+1: thanks Tim!

spec/stream_spec.rb
```diff
@@ -56,13 +56,15 @@
...
+          expect { @stream.dup.send type }.to raise_error InternalError
         end
       end
 
       it 'should raise error on receipt of invalid frames' do
-        (FRAME_TYPES - [PRIORITY, RST_STREAM, WINDOW_UPDATE]).each do |type|
-          expect { @stream.dup.receive type }.to raise_error StreamError
+        what_types = (FRAME_TYPES - [PRIORITY, RST_STREAM, WINDOW_UPDATE])
+        what_types.each do |type|
+          p "test type #{type}"
```
mad-p: Debug print remained

mad-p:
Thanks, Tim! lgtm

PR #39, opened 2015-08-20
Corrects #37.

tbetbetbe:
The reference to the flags array that will be shared between successive chunks (before they are actually emitted); this can cause streams to terminate early.

### DIFF
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -173,16 +173,15 @@ module HTTP2
     # @param payload [String]
     # @param end_stream [Boolean] indicates last response DATA frame
     def data(payload, end_stream: true)
-      flags = []
-
       # Split data according to each frame is smaller enough
       # TODO: consider padding?
       max_size = @connection.remote_settings[:settings_max_frame_size]
       while payload.bytesize > max_size
         chunk = payload.slice!(0, max_size)
-        send(type: :data, flags: flags, payload: chunk)
+        send(type: :data, flags: [], payload: chunk)
       end
 
+      flags = []
       flags << :end_stream if end_stream
       send(type: :data, flags: flags, payload: payload)
     end


lib/http/2/stream.rb
```diff
@@ -173,16 +173,15 @@ def reprioritize(weight: 16, dependency: 0, exclusive: false)
...
     def data(payload, end_stream: true)
-      flags = []
-
       # Split data according to each frame is smaller enough
       # TODO: consider padding?
       max_size = @connection.remote_settings[:settings_max_frame_size]
       while payload.bytesize > max_size
         chunk = payload.slice!(0, max_size)
-        send(type: :data, flags: flags, payload: chunk)
+        send(type: :data, flags: [], payload: chunk)
```
igrigorik: Doh, right, should have caught that in the previous pull. 

DATA frames can also be padded, so if we're splitting a large payload we should probably pass the flag through (if specified) and pad each frame. That said, we don't actually support padding today, so that's a bit of a noop. I guess we  already have the TODO above, so it's ok as is.

lib/http/2/stream.rb
```diff
@@ -173,16 +173,15 @@ def reprioritize(weight: 16, dependency: 0, exclusive: false)
     # @param payload [String]
     # @param end_stream [Boolean] indicates last response DATA frame
     def data(payload, end_stream: true)
-      flags = []
```
igrigorik: Minor thing: can we leave this declaration up here and pass through flags (which is set to empty array)? The benefit is that we don't end up allocating a new object on each iteration of the while loop.

lib/http/2/stream.rb
tbetbetbe: Just checking then: Do I need to anything more on this PR?

igrigorik: One nitpick: https://github.com/igrigorik/http-2/pull/39/files#r37494623

PR #43, opened 2015-12-27
Add nils to initialize to prevent syntax error

dfried:
There seems to be an issue with 0.8.0 where you can't initialize a stream. I think it's a simple syntax error and this fixed it.

### DIFF
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -71,7 +71,7 @@ module HTTP2
     # @param exclusive [Boolean]
     # @param window [Integer]
     # @param parent [Stream]
-    def initialize(connection:, id:, weight: 16, dependency: 0, exclusive: false, parent: nil)
+    def initialize(connection: nil, id: nil, weight: 16, dependency: 0, exclusive: false, parent: nil)
       @connection = connection
       @id = id
       @weight = weight


igrigorik:
Hmm, are you referring to the [rbx build](https://travis-ci.org/igrigorik/http-2)?

dfried:
No, I was getting a syntax error on the first two args having colons at the end.

igrigorik:
Which Ruby version? All the specs pass here and on Travis?

alloy:
This is because the gemspec only requires Ruby 2.0.0, but keyword arguments without default values weren’t supported before 2.1.0.

I have a change to the gem spec patch for this https://github.com/alloy/http-2/commit/74f387e944d84c8e78eb28e694716ead5a5a489d, I added it as part of my #44 PR. I should revert that if you choose to fix it by adding default values, as this PR does.

igrigorik:
Are you aware of any stats on 2.0.0 use/adoption vs newer versions? Ruby site indicates that 2.0.0 will EOL'ed soon, but with no particular date.

/cc @mad-p wdyt?

mkonecny:
Why not merge this PR and support a much wider range of Ruby installations? The end-result is exactly the same. 

Our systems are currently on Ubuntu 14.04 (the latest LTE) with Ruby 2.0, which makes it a very common setup.

igrigorik:
https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

> As it has been announced before, all support for Ruby 2.0.0 has ended today. Bug and security fixes from more recent Ruby versions will no longer be backported to 2.0.0, and no further patch release of 2.0.0 will be released.

We don't want to promote use of 2.0.0.

PR #44, opened 2015-12-29
[client] Be sure to send the preface if receiving data first.

alloy:
I’m very new to HTTP/2 and this library, so I’m mostly looking for guidance as to why this is needed and where to make the changes in the codebase.

I’m using this lib to connect to Apple’s new Push Notification service. Without this change it would complain about missing a client preface and instead receiving other data (which were settings). What would happen is that the server would send data before I would send any, which would change the `@state` to connected and thus the preface would never get send from `Client#send`.

### DIFF
--- a/lib/http/2/client.rb
+++ b/lib/http/2/client.rb
@@ -34,13 +34,23 @@ module HTTP2
     # @see Connection
     # @param frame [Hash]
     def send(frame)
-      send_connection_preface
+      if @state == :waiting_connection_preface
+        send_connection_preface
+      end
       super(frame)
+    end
+
+    def connection_management(frame)
+      if @state == :waiting_connection_preface
+        send_connection_preface
+        connection_settings(frame)
+      else
+        super(frame)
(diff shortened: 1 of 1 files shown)


alloy:
Any feedback for me to wrap up this PR?

igrigorik:
> What would happen is that the server would send data before I would send any, which would change the @state to connected and thus the preface would never get send from Client#send.

As in, you open a TCP connection and the server immediately sends a preface? If so, that's really weird.. I'm not familiar with Apple's implementation, but if that's true then it would imply that they treat their end as the client, even though they don't initiate the connection?

igrigorik:
Any additional insights or progress on this one?

alloy:
Sorry, missed your earlier response. Based on that error message it sends, I think Apple just uses an existing Java server. I’ll do some re-checking of what goes on and provide better logs.

igrigorik:
@alloy any luck with this one?

PR #51, opened 2016-02-27
Run hpack compressor test with mode

kiennt:
This patch fix a minor bug when run compressor specs: It changes options in Compressor and Decompressor so they will run with options all of encoding options


igrigorik:
Thank you!

PR #53, opened 2016-02-29
play nice with older rubies

slact:
This fixes a syntax error for ruby 2.0

### DIFF
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -71,7 +71,7 @@ module HTTP2
     # @param exclusive [Boolean]
     # @param window [Integer]
     # @param parent [Stream]
-    def initialize(connection:, id:, weight: 16, dependency: 0, exclusive: false, parent: nil)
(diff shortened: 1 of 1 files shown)


igrigorik:
https://github.com/igrigorik/http-2/blob/master/http-2.gemspec#L15 < this is a no-op, I think.

igrigorik:
From: https://www.ruby-lang.org/en/news/2016/02/24/support-plan-of-ruby-2-0-0-and-2-1/

> As it has been announced before, all support for Ruby 2.0.0 has ended today. Bug and security fixes from more recent Ruby versions will no longer be backported to 2.0.0, and no further patch release of 2.0.0 will be released. We highly recommend that you upgrade to Ruby 2.3 or 2.2 as soon as possible.

With above in mind, closing this, as it encourages people to migrate to 2.2+.

slact:
ok, no problem, that makes sense.

Issue #58, opened 2016-10-06
Documentation for the promise?


georgeu2000:
Thank you for all your awesome work on HTTP2. 

The following code from the docs seems strange:
```
  head = {
     ":status" => 200,
     ":path"   => "/other_resource",
     "content-type" => "text/plain"
    }

    # initiate server push stream
    stream.promise(head) do |push|
      push.headers({ ... })
      push.data(...)
    end

    # send response
    stream.headers({
      ":status" => 200,
      "content-type" => "text/plain"
    })

    # split response between multiple DATA frames
    stream.data(response_chunk, end_stream: false)
    promise.data(payload)
    stream.data(last_chunk)
```

It's not clear what data would be pushed in the promise. According to [HPBN](https://hpbn.co/http2/#push-promise-101)

> PUSH_PROMISE frames, which signal the server’s intent to push the described resources to the client and need to be delivered ahead of the response data that requests the pushed resources.

What data would be sent in the PUSH_PROMISE if the data is supposed to be sent later?

Also, `promise` is not defined at this point: `promise.data(payload)`. Is this supposed to refer to the promise stream? If so, how to access it? Or are the promise and the response stream the same? In that case, would it be `stream.data(promise_data)`?

igrigorik:
> What data would be sent in the PUSH_PROMISE if the data is supposed to be sent later?

PUSH_PROMISE delivers headers meta-data about the resource you're promising to push, thus notifying the client that if it wants to fetch this resource.. it should hold-on because the server is already planning to send it this data (when exactly is at the discretion of the server). 

> Also, promise is not defined at this point: promise.data(payload). Is this supposed to refer to the promise stream? If so, how to access it? 

Yes, you're right that's a bug. It should refer to the "push" stream established above.

> Or are the promise and the response stream the same? In that case, would it be stream.data(promise_data)?

No, they're *not* the same stream. When you call .promise you get back a new stream: https://github.com/igrigorik/http-2/blob/master/lib/http/2/server.rb#L118

georgeu2000:
OK, thanks for the info.

How can I contribute?

igrigorik:
@georgeu2000 any and all code/documentation/example patches (pull requests) would be most welcome! I could have sworn I had a push example lying around.. but I think I accidentally wiped it before I committed it. Having a working h2 push example would be a great contribution.. if you're up for it :)

PR #60, opened 2016-10-08
update readme for promise and server push

georgeu2000:
Issue https://github.com/igrigorik/http-2/issues/58

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,9 +246,10 @@ conn.on(:stream) do |stream|
     }
 
     # initiate server push stream
+    push_stream = nil
     stream.promise(head) do |push|
-      push.headers({ ... })
-      push.data(...)
+      push.headers(head)
+      push_stream = push
     end
 
     # send response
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -246,9 +246,10 @@ conn.on(:stream) do |stream|
     }
 
     # initiate server push stream
+    push_stream = nil
     stream.promise(head) do |push|
-      push.headers({ ... })
-      push.data(...)
```
georgeu2000: It seems like we should send the data later.

igrigorik: That's what .promise does above.. The latter call is to demonstrate that you can send additional headers if you want, before you proceed with body.

igrigorik:
LGTM, modulo the .headers({...}) change; we shouldn't send same headers data twice.

README.md
georgeu2000: In my testing, this sends the data immediately. I don't see how it sends it later.

```ruby
stream.promise(head) do |push|
       push.headers({ ... })
      push.data(...)
```

igrigorik: Sorry about the confusion, I meant that we should keep .headers({...}) in place to demonstrate how to send additional headers. the .data() calls can happen in the callback or later -- this is up to the application to decide based on what it's pushing, etc.

georgeu2000: Oh, I see.

igrigorik:
Scratching my head as to why Travis is failing all of the sudden.. but we'll deal with that separately.

Looks good now, thanks!

PR #61, opened 2016-10-08
server push example

georgeu2000:


example/client.rb
```diff
@@ -68,7 +68,9 @@
 
 stream.on(:close) do
   log.info 'stream closed'
-  sock.close
+
+  # keep connection open to handle push messages
+  # sock.close
```
igrigorik: No, this is not correct. The parent stream must be open for PUSH_PROMISE to be valid; the server can't arbitrarily push against closed and/or non-existent streams.

example/helper.rb
```diff
@@ -17,3 +17,25 @@ def info(msg)
...
+    puts 'sending push'
+
+    head = {
+      ':status' => '200',
+      ':path'   => "/other_resource/#{i}",
+      'content-type' => 'text/plain',
+    }
+
+    stream.promise(head) do |push|
+      push.headers(head)
```
igrigorik: you don't need to send this twice.

example/helper.rb
```diff
@@ -17,3 +17,25 @@ def info(msg)
     puts "[Stream #{@id}]: #{msg}"
   end
 end
+
+def send_promises(stream)
```
igrigorik: Why keep this in helper? It's only applicable to server.rb, no?

georgeu2000: Yes, that's true. It is only for the server.

Not sure what would be a good place to put it, unless we have a `server_helper`. Putting it at the top of `server.rb` would not be good...

example/client.rb
georgeu2000: I think that what I did. Keep the stream open.

example/helper.rb
georgeu2000: So, I can create a `server_helper`, if you think that makes sense.

example/helper.rb
georgeu2000: Without `push.headers(head)`, sending data raises an `InternalError`.

```
Sent frame: {:type=>:push_promise, :flags=>[:end_headers], :stream=>1, 
:promise_stream=>2, :payload=>[[":status", "200"], [":path", "/other_resource/0"], 
    ["content-type", "text/plain"]]}

Sent frame: {:type=>:headers, :flags=>[:end_headers],
 :payload=>[[":status", "200"], [":path", "/other_resource/0"], ["content-type", "text/plain"]],
 :stream=>2}
````

It seems that `stream.promise(head)` sends the promise on the "original" stream. `push.headers(head)` appears to send the headers on the "push" stream.

File: example/client.rb (hunk 1 above)
Comment (igrigorik) [inline example/client.rb]: :close callback is called when the other end closes the stream.. At this point, we can't push against this stream anymore. Keep the sock.close as is :)

File: example/helper.rb (hunk 2 above)
Comment (igrigorik) [inline example/helper.rb]: I guess this one is a bit subtle.. We may want to rethink the API we provide here in the future, but for time being:

http://httpwg.org/specs/rfc7540.html#PushResources

> The PUSH_PROMISE frame includes a header block that contains a complete set of **request header fields** that the server attributes to the request.
> 
> The header fields in PUSH_PROMISE and any subsequent CONTINUATION frames MUST be a valid and complete set of request header fields (Section 8.1.2.3). The server MUST include a method in the :method pseudo-header field that is safe and cacheable.
>
> The server MUST include a value in the :authority pseudo-header field for which the server is authoritative (see Section 10.1). A client MUST treat a PUSH_PROMISE for which the server is not authoritative as a stream error (Section 5.4.2) of type PROTOCOL_ERROR.

So, what we need to do here is:

```ruby
stream.promise({':method': 'GET', ':authority': 'localhost', ':path': '/resource/...'}) do |push|
  push.headers({':status': 200, 'content-length': ..., ... })
  push.data(...)
end
```

Hope that makes sense!

example/helper.rb
georgeu2000: OK, I'll update it.

example/client.rb
georgeu2000: The push example uses multiple streams. So, the socket should stay open until all pushes finish. This seemed like the simplest way to do it.

example/helper.rb
georgeu2000: Thank you for the explanation. Code is updated.

example/helper.rb
georgeu2000: Moved to `server.rb`.

example/client.rb
igrigorik: OK, I guess that's fine -- it's just a demo server. Can you drop the comments though? My issue is what the message indirectly implies: just because the connection is open doesn't mean you can handle push messages.

If we drop both lines then we're good.

example/server.rb
```diff
@@ -88,9 +92,36 @@
...
+
+        # send 10 promises
+        10.times do |i|
+          sleep 1
+          puts 'sending push'
+
+          head = { ':method' => 'GET',
+                   ':authority' => 'localhost',
+                   ':path' => "/other_resource/#{i}",
+                   'content-type' => 'text/plain' }
```
igrigorik: content-type is a response header, it should move down.. or, we can just drop it.

igrigorik:
Awesome, thanks for the updates! Almost there... just a few more small nitpicks. :)

example/client.rb
georgeu2000: Sure, no problem.

igrigorik:
👍  awesome work, and thanks for your patience! :)

georgeu2000:
My pleasure.

PR #62, opened 2016-10-11
Fix ci

georgeu2000:
This [commit](https://github.com/igrigorik/http-2/commit/31b7b20328ef2dd28427d76af1865a409eea2e88) introduced ActiveSupport to use `#deep dup`. This created a dependency on Ruby 2.2.2, which is causing the CI to fail.

Not sure if we need to support previous versions of Ruby, but it doesn't seem necessary to include Active Support for a few methods.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ gem 'rake'
 gem 'yard'
 
 group :test do
-  gem 'activesupport'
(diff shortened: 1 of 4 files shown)


igrigorik:
Coveralls has gone haywire.. :)

LGTM.

PR #65, opened 2016-10-12
buffer wraps string

georgeu2000:
> TODO: Refactor, it would be better if Buffer were not a String subclass,
>  but rather wrap a string and only expose the mutating API needed so that
>   the possible surface for things to go wrong stays controllable.
>   [Pull #46](https://github.com/igrigorik/http-2/pull/46)

### DIFF
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -149,9 +149,7 @@ RSpec.describe HTTP2::Connection do
       ]
       headers = []
       @conn.on(:frame) do |bytes|
-        # buf[3]: frame's type field
-        buf = Buffer.new(bytes)
-        headers << f.parse(buf) if [1, 5, 9].include?(buf[3].ord)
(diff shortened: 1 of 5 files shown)


spec/connection_spec.rb
```diff
@@ -149,8 +149,9 @@
       ]
       headers = []
       @conn.on(:frame) do |bytes|
-        # bytes[3]: frame's type field
-        headers << f.parse(bytes) if [1, 5, 9].include?(bytes[3].ord)
+        # buf[3]: frame's type field
+        buf = Buffer.new(bytes)
```
igrigorik: Perhaps we should pass in a buffer object in the callback? Manually converting to a buffer each time seems like a gotcha / unnecessary complication.

georgeu2000: Turns out it is not necessary.

igrigorik:
👍

georgeu2000:
My pleasure.

PR #66, opened 2016-10-18
be safe with nils in the connection_error handler

kenichi:
ran into a few states where `msg` _and_ `e` were `nil`


kenichi:
eeep - i know they're >= 2.3 only 😳   maybe we do something different

igrigorik:
Yeah, let's expand it to the longer version.. (e && e.message), etc.

georgeu2000:
This is much cleaner than my code. Shall we use this instead?

igrigorik:
sgtm

kenichi:
ok  :)

i rebased, so i had to open #84

PR #68, opened 2016-10-23
Fix examples

georgeu2000:


igrigorik:
> We really should have some tests to keep these examples working. ..

Indeed, that would be great.

igrigorik:
Thanks for fixing these!

georgeu2000:
My pleasure.

PR #74, opened 2016-11-15
Fix connection_error when Exception is NoMethodError or nil

georgeu2000:
When working on [h2spec](https://github.com/summerwind/h2spec), discovered that connection_error exception can be `NoMethodError` or `nil`. Calling `e.message` or `e.backtrace` raises a new exception.

Not sure why the exception would be `NoMethodError` or `nil` (seems strange to me), but this fixes it for now.

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -654,7 +654,14 @@ module HTTP2
 
       @state, @error = :closed, error
       klass = error.to_s.split('_').map(&:capitalize).join
-      fail Error.const_get(klass), msg || e.message, e.backtrace || []
+
+      message = msg
+      message = e.message if message.nil? && e.respond_to?(:message)
+
+      backtrace = []
+      backtrace = e.backtrace if e.respond_to?(:backtrace)
+
+      fail Error.const_get(klass), message, backtrace
     end
   end
 end


igrigorik:
Do you have an example backtrace of when NoMethodError is triggered?

georgeu2000:
After looking into it, I see that the NoMethodError was being raised from a typo that I introduced. It has nothing to do with http-2.

igrigorik:
:)

So, the question is.. in such cases, what should http-2 raise? We should catch it and goaway the connection, etc... I think.

georgeu2000:
The connection is already goaway'd. This is an issue **_within_** the error handler.

```
  def connection_error(error = :protocol_error, msg: nil, e: nil)
    goaway(error) unless @state == :closed || @state == :new

    @state, @error = :closed, error
    klass = error.to_s.split('_').map(&:capitalize).join
    fail Error.const_get(klass), msg || e.message, e.backtrace || []
  end
```

```
5.1. Stream States
     × half closed (remote): Sends a DATA frame
      - The endpoint MUST respond with a stream error (Section 5.4.2) of type STREAM_CLOSED.
        Expected: GOAWAY frame (ErrorCode: STREAM_CLOSED)
                  RST_STREAM frame (ErrorCode: STREAM_CLOSED)
                  Connection close
          Actual: WINDOW_UPDATE frame (Length: 4, Flags: 0)
```

Even with the fix, the above spec fails. I think that receive does not handle this case because it is raising the connection_error at the [end of the method](https://github.com/igrigorik/http-2/blob/master/lib/http/2/connection.rb#L322).

```
# Decodes incoming bytes into HTTP 2.0 frames and routes them to
    # appropriate receivers: connection frames are handled directly, and
    # stream frames are passed to appropriate stream objects.
    #
    # @param data [String] Binary encoded string
    def receive(data)
```

It seems to me that fixing h2spec 5.1 is not really part of this issue.

igrigorik:
Ok, sounds like that's a separate investigation then. Is this good to merge from your perspective?

georgeu2000:
Yes.

kenichi:
#66 ?

PR #76, opened 2016-11-16
requiring base64 (Closes #75)

### LINKED ISSUE #75: NameError: uninitialized constant HTTP2::Server::Base64
Here: https://github.com/igrigorik/http-2/blob/a3298edb2bd20cc3821536509c99634fe0425aa6/lib/http/2/server.rb#L74

I guess this has only been tested with the upgrade_server.rb example, as the require is being done there. 

You can require the base64 lib in the server file and be done with it. However, you can also get away by unpacking the string, thereby bypassing require:

```ruby
require "base64"
Base64.decode64(base64str)
# equals
base64str.unpack("m*")[0]
```

TiagoCardoso1983:
@igrigorik as you are using the urlsafe_ version of the decoding, I don't know if it's right to unpack the string myself. So I just required the module. Tell me what you think. The `Base64` is anyway just a small set of wrapper methods around string unpacking.

### DIFF
--- a/example/upgrade_server.rb
+++ b/example/upgrade_server.rb
@@ -2,7 +2,6 @@
 
 require_relative 'helper'
 require 'http_parser'
-require 'base64'
 
 options = { port: 8080 }
 OptionParser.new do |opts|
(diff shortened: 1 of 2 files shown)


igrigorik:
👍 thanks!

Issue #72, opened 2016-11-08
Update core + example server/client to pass h2spec


igrigorik:
Instructions to build and run: https://github.com/summerwind/h2spec

Running `./h2spec -p 8080` against our example/server.rb yield lots bunch of failures:

- example server aborts the connection instead of sending GOAWAY / PROTOCOL_ERROR in bunch of places
- we don't enforce/emit FRAME_SIZE_ERROR 
- ...

In short, lots of opportunities to improve error feedback!

georgeu2000:
h2spec doesn't even run against insecure server for me(see below). 

However, it does run against the secure server with 73 tests, 33 passed, 0 skipped, 40 failed.

I'll look into it...

```
go-workspace> ./bin/h2spec -p 8080
  3.5. HTTP/2 Connection Preface
    ✓ Sends invalid connection preface

  4.2. Frame Size
      Sends large size frame that exceeds the SETTINGS_MAX_FRAME_SIZE
```

```
HTTP2::Error::HandshakeError exception: HTTP2::Error::HandshakeError - closing socket.
	/Users/george/repos/http-2/lib/http/2/connection.rb:192:in `receive'
	example/server.rb:134:in `block in <main>'
	example/server.rb:46:in `loop'
	example/server.rb:46:in `<main>'
```

igrigorik:
@georgeu2000 guessing, our demo server keels over and stops accepting connections..

georgeu2000:
@igrigorik - I think what is happening is that we are receiving a frame of length 0:

    Received frame: {:length=>0, :type=>:settings, :flags=>[], :stream=>0, :payload=>[]}

Which is strange because the spec is

    4.2. Frame Size
      Sends large size frame that exceeds the SETTINGS_MAX_FRAME_SIZE

The buffer length is 0 and Framer is returning nil ([Framer](https://github.com/igrigorik/http-2/blob/master/lib/http/2/framer.rb#L322-L327)):

    # Decodes complete HTTP/2 frame from provided buffer. If the buffer
    # does not contain enough data, no further work is performed.
    #
    # @param buf [Buffer]
    def parse(buf)
      return nil if buf.size < 9

However the connection is waiting for a frame ([Connection](https://github.com/igrigorik/http-2/blob/master/lib/http/2/connection.rb#L196)): 

      while (frame = @framer.parse(@recv_buffer))

So there is no response. Not sure what the desired functionality is yet, but perhaps it should emit a Settings frame...?

[...]

PR #77, opened 2016-11-16
Fix non-secure server to run against h2spec

georgeu2000:
Does not seem to send any data on TCP Socket. Allows server.rb to run against h2spec.

Issue https://github.com/igrigorik/http-2/issues/72#issuecomment-261000121

### DIFF
--- a/example/server.rb
+++ b/example/server.rb
@@ -50,7 +50,11 @@ loop do
   conn = HTTP2::Server.new
   conn.on(:frame) do |bytes|
     # puts "Writing bytes: #{bytes.unpack("H*").first}"
-    sock.write bytes
+    if sock.is_a?(TCPSocket)
+      sock.send bytes, 0
+    else
+      sock.write bytes
+    end
   end
   conn.on(:frame_sent) do |frame|
     puts "Sent frame: #{frame.inspect}"
--- a/example/server.rb
+++ b/example/server.rb
@@ -50,11 +50,7 @@ loop do
   conn = HTTP2::Server.new
   conn.on(:frame) do |bytes|
     # puts "Writing bytes: #{bytes.unpack("H*").first}"
-    if sock.is_a?(TCPSocket)
-      sock.send bytes, 0
-    else
-      sock.write bytes
-    end
+    sock.is_a?(TCPSocket) ? sock.sendmsg(bytes) : sock.write(bytes)
   end
   conn.on(:frame_sent) do |frame|
     puts "Sent frame: #{frame.inspect}"


igrigorik:
Hmm, the flags are a bit hard to reason about.. Instead, should we consider setting the NODELAY option instead? Can you give it a try and see if it fixes the problem?

`sock.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1)` should, in theory, do the trick when we initialize the socket.

georgeu2000:
Yes, that works. I'll update.

georgeu2000:
Sorry, actually `setsockopt` doesn't work. (Still had `sock.send bytes, 0`). I'll can into `setsockopt` and see if I can figure out what is going on...

georgeu2000:
Looked into `Socket::IPPROTO_TCP, Socket::TCP_NODELAY` and socket delay does not seem to be the issue.

`sock.sendmsg bytes` works as well and does not expose the flags, so would that be OK?

`sock.sendmsg bytes` sets the MSG_ flags to 0, which apparently does something besides disabling socket delay.

```
MSG_COMPAT
MSG_CONFIRM
MSG_CTRUNC
MSG_DONTROUTE
MSG_DONTWAIT
MSG_EOF
MSG_EOR
MSG_ERRQUEUE
MSG_FIN
MSG_FLUSH
MSG_HAVEMORE
MSG_HOLD
MSG_MORE
MSG_NOSIGNAL
MSG_OOB
MSG_PEEK
MSG_PROXY
MSG_RCVMORE
MSG_RST
MSG_SEND
MSG_SYN
MSG_TRUNC
MSG_WAITALL
```

example/server.rb
```diff
@@ -50,7 +50,11 @@
   conn = HTTP2::Server.new
   conn.on(:frame) do |bytes|
     # puts "Writing bytes: #{bytes.unpack("H*").first}"
-    sock.write bytes
+    if sock.is_a?(TCPSocket)
+      sock.send bytes, 0
```
igrigorik: It's a little weird to use send in one and write in another. I believe send is just an alias.. can we use write? Related, do we actually need the if check here?

igrigorik:
Interesting... =\

I guess I'm good either way. See my question above.

example/server.rb
georgeu2000: That's the whole issue. `IO#write` works differently from `TCPSocket#sendmsg`. I haven't figured out what the difference is.

Here is the hierarchy: `IO < BasicSocket < IPSocket < TCPSocket`.

In contrast, `OpenSSL::SSL::SSLSocket` does not inherit from any of the above, and the #write method works differently.

Yes, we need to check the class because `#sendmsg` does not work with SSLSocket.

I think we should create a Server model from the server.rb file, and then a method would encapsulate the `#write` and `#sendmsg` methods.

PR #81, opened 2016-11-18
Create rake task to run h2spec

georgeu2000:
`rake h2spec`

Limitations: 
1. Only works on OSX.
2. Only works for non-secure server. Need to figure out how to manage output because it is already too long.

Since the output is long (600 lines), I think we can leave it out of the default task. Alternatively, we could abbreviate the result to `73 tests, 33 passed, 0 skipped, 40 failed`. 

One thing we could do is append the truncated output to the default rake task, but provide full output for `rake h2spec`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,6 +12,15 @@ RSpec::Core::RakeTask.new(:hpack) do |t|
   t.pattern = './spec/hpack_test_spec.rb'
 end
 
+task :h2spec do
+  system 'ruby example/server.rb -p 9000 &', out: File::NULL
+  sleep 1
+
+  system 'spec/h2spec/h2spec.darwin -p 9000'
+
+  system 'kill `pgrep -f example/server.rb`'
+end
+
 RuboCop::RakeTask.new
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -12,6 +12,15 @@ RSpec::Core::RakeTask.new(:hpack) do |t|
   t.pattern = './spec/hpack_test_spec.rb'
 end
 
+task :h2spec do
+  system 'ruby example/server.rb -p 9000 &', out: File::NULL
+  sleep 1
+
+  system 'spec/h2spec/h2spec.darwin -p 9000'
```
igrigorik: At a minimum, we should provide a comment for where to download the binary?

We can also exit early with an error for non-darwin:```if /darwin/ !~ RUBY_PLATFORM) ... ```

georgeu2000:
Forgot to commit binary. Done.

Added check for OSX.

igrigorik:
Looks like rubocop is not happy, can you fix the comment? :)

igrigorik:
👍

PR #83, opened 2016-11-21
h2spec fix 5.1.1. Stream Identifiers

georgeu2000:
× Sends even-numbered stream identifier
        - The endpoint MUST respond with a connection error of type PROTOCOL_ERROR.

73 tests, 33 passed, 0 skipped, 40 failed

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -226,6 +226,12 @@ module HTTP2
         else
           case frame[:type]
           when :headers
+            # When even-numbered stream identifier is received,
+            # the endpoint MUST respond with a connection error of type PROTOCOL_ERROR.
+            if frame[ :stream ].even?
+              connection_error
+            end
+
             # The last frame in a sequence of HEADERS/CONTINUATION
             # frames MUST have the END_HEADERS flag set.
             unless frame[:flags].include? :end_headers
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -226,11 +226,9 @@ module HTTP2
         else
           case frame[:type]
           when :headers
-            # When even-numbered stream identifier is received,
+            # When server receives even-numbered stream identifier,
             # the endpoint MUST respond with a connection error of type PROTOCOL_ERROR.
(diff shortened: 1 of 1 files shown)


lib/http/2/connection.rb
```diff
@@ -226,6 +226,12 @@ def receive(data)
         else
           case frame[:type]
           when :headers
+            # When even-numbered stream identifier is received,
+            # the endpoint MUST respond with a connection error of type PROTOCOL_ERROR.
+            if frame[ :stream ].even?
```
igrigorik: Err, wouldn't this break client apps? As in, connection.rb is used by both client and server classes, and this check effectively forces client-only code path for header processing.

georgeu2000:
Sorry, I forgot to run the http-2 specs.

I made an update but wasn't sure what you meant by

> this check effectively forces client-only code path for header processing

georgeu2000:
BTW, it's difficult to track how many and which h2specs are passing and failing. I think we need a formatter to concisely list the failures...

lib/http/2/connection.rb
```diff
@@ -226,6 +226,10 @@ def receive(data)
         else
           case frame[:type]
           when :headers
+            # When server receives even-numbered stream identifier,
+            # the endpoint MUST respond with a connection error of type PROTOCOL_ERROR.
+            connection_error if frame[:stream].even? && is_a?(Server)
```
igrigorik: 👍 

Minor nitpick (surprised rubocop didn't complain): https://github.com/georgeu2000/http-2/blob/f52efc656619f5f05b4f69e203f2ffd1ae4bd10e/lib/http/2/connection.rb#L638 - for consistency, can we self.is_a?

georgeu2000: Sure.

georgeu2000:
Regarding h2spec, perhaps a better solution is to save the output to a file. Then a diff will show what is fixed and what is broken.

georgeu2000:
@igrigorik - Haven't heard back so I guess it is OK. Merging...

igrigorik:
👍

PR #87, opened 2016-11-28
Update h2spec output for Only connection error if connection is a server

georgeu2000:


georgeu2000:
BTW h2spec 5.5 and 6.2 are failing intermittently.

PR #89, opened 2016-12-13
Use ALPN instead of NPN

mkauf:
The HTTP/2 specification suggests to use ALPN for HTTP/2 over TLS.

ALPN is supported by Ruby 2.3 and newer. Maybe someone can improve this pull request, so that the code still runs with older Ruby versions.


igrigorik:
+1 for using ALPN. 

This is a example server, I think we can just add a note that 2.3.1 or higher is needed for the ALPN callback. Would you mind adding a comment in the code for this?

mkauf:
Good idea, I have added a comment in the code.

igrigorik:
Great stuff, thanks!

PR #90, opened 2016-12-13
Buffer: also forward the to_s method

mkauf:
Instead of actual data, the example program "client.rb" sent the string ``"#<HTTP2::Buffer:0x0056529cd1d1a8>"`` when I tried it. Probably this behavior depends on the Ruby version (I have tested with Ruby 2.3.1).

I don't know if there is a better bugfix, but also forwarding the "to_s" method fixes the problem.

### DIFF
--- a/lib/http/2/buffer.rb
+++ b/lib/http/2/buffer.rb
@@ -7,7 +7,7 @@ module HTTP2
     extend Forwardable
 
     def_delegators :@buffer, :ord, :encoding, :setbyte, :unpack,
-                   :size, :each_byte,  :to_str, :length, :inspect,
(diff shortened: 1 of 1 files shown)


igrigorik:
Some context on to_s vs to_str (which we already implement): https://gist.github.com/kinopyo/5682347

@georgeu2000 WDYT? I'll defer to you on this one.

georgeu2000:
@igrigorik Regarding context link:

> You should not implement to_str unless your object acts like a string

It seems to me that a buffer **_does_** act like a string, so it should implement `to_str`.

LGTM. Merging.

mkauf:
Thank you for merging!

PR #91, opened 2016-12-16
Update README.md

msroot:
fix `ruby server.rb --push`
`-p` its a reserved flag for `port`

```
$ ruby server.rb -push
Starting server on port ush
server.rb:21:in `initialize': getaddrinfo: nodename nor servname provided, or not known (SocketError)
```


igrigorik:
Good catch, thanks!

Issue #97, opened 2017-03-20
Header Compression Issue (COMPRESSION_ERROR)


HoneyryderChuck:
I'm having a hard-to-reproduce issue which has to do with the state of the headers in the browser, but not in other clients. Namely, the browser isn't being able to decompress the headers after a certain point in an http2 connection/session. Specifically, I'm receiving a `goaway` frame from the browser with `COMPRESSION_ERROR` type and `"Framer error: 6 (DECOMPRESS_FAILURE)"`. The browser is Chrome  56.0.2924.87 (64-bit), and this has been tested against the master branch of `http-2`. 

The server log is the following (only logging the goaway as it happens):

```
 - - - [20/Mar/2017:15:17:27 +0100] "GET / 2.0" 200 - 1.7691
- - - [20/Mar/2017:15:17:27 +0100] "GET /leaflet/dist/leaflet.css?body=1 2.0" 200 10162 0.0298
- - - [20/Mar/2017:15:17:27 +0100] "GET /cart.css?body=1 2.0" 200 4186 0.0167
- - - [20/Mar/2017:15:17:27 +0100] "GET /auth_forms.css?body=1 2.0" 200 1874 0.0327
- - - [20/Mar/2017:15:17:27 +0100] "GET /application.css?body=1 2.0" 200 874 0.0791
- - - [20/Mar/2017:15:17:27 +0100] "GET /home.css?body=1 2.0" 200 365445 0.0978
- - - [20/Mar/2017:15:17:27 +0100] "GET /orders.css?body=1 2.0" 200 119 0.0924
- - - [20/Mar/2017:15:17:27 +0100] "GET /pizzas.css?body=1 2.0" 200 - 0.1002
- - - [20/Mar/2017:15:17:27 +0100] "GET /registrations.css?body=1 2.0" 200 - 0.0689
- - - [20/Mar/2017:15:17:27 +0100] "GET /jquery.js?body=1 2.0" 200 293431 0.0660
- - - [20/Mar/2017:15:17:27 +0100] "GET /jquery_ujs.js?body=1 2.0" 200 21600 0.0658
- - - [20/Mar/2017:15:17:27 +0100] "GET /bootstrap.js?body=1 2.0" 200 120 0.0673
- - - [20/Mar/2017:15:17:27 +0100] "GET /bootstrap/alert.js?body=1 2.0" 200 2261 0.0666
[...]

PR #98, opened 2017-03-24
Table Size updates

HoneyryderChuck:
@igrigorik fwiw I'm still not confident about this, but it was the minimal change that didn't break the specs (as stated previously, maybe false positive) and allowed my tests to succeed. Maybe you can help me track this. 

I'm afraid that, by switching those, I might have broken the client implementation. I've seen a few specs playing with the header table size value, and they're not breaking, so I don't know what to think.


igrigorik:
LGTM: https://github.com/igrigorik/http-2/issues/97#issuecomment-289229312

PR #104, opened 2017-10-09
Replace deprecated tmp_ecdh_callback with ecdh_curves

southwolf:
`tmp_ecdh_callback=` is deprecated [Note](https://github.com/ruby/openssl/blob/master/History.md)
Replace it with `ecdh_curves= 'P-256'`


igrigorik:
Thanks for the report and fix 👍

PR #107, opened 2017-10-23
fixed window update by adding receive window control

HoneyryderChuck:
as it was, a WINDOW_UPDATE frame was being send for every DATA frame received; this
generated a flood of control frames for small DATA frames sent at a
time, downgrading performance; implemented receive window strategy,
which implements a similar heuristic as nghttp, where the receiver keeps
the size of DATA payload received - when this goes above half of the
local window, the receiver issues a WINDOW_UPDATE with the receiver
window as increment.

for doc:

* https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4941-L4958


HoneyryderChuck:
@igrigorik any feedback? FWIW this solves my timeout issues, and behaves closer to what nghttp does.

lib/http/2/stream.rb
```diff
@@ -78,6 +78,7 @@ def initialize(connection:, id:, weight: 16, dependency: 0, exclusive: false, pa
       @dependency = dependency
       process_priority(weight: weight, stream_dependency: dependency, exclusive: exclusive)
       @local_window  = connection.local_settings[:settings_initial_window_size]
+      @recv_window_size = 0
```
igrigorik: Why do you need this? Aren't you simply counting the exact same thing twice, in one case by decrementing from @local_window and in another by incrementing @recv_window_size?

lib/http/2/stream.rb
```diff
@@ -587,6 +592,13 @@ def process_priority(frame)
       #   We should add code here.
     end
 
+    def adjust_recv_window_size(delta_size)
+      if @recv_window_size > @local_window || @recv_window_size > 0x7fffffff - delta_size
+        stream_error(:flow_control_error)
```
igrigorik: Where did case come from? It's not tested and it's not clear how or why this would be triggered?

lib/http/2/stream.rb
```diff
@@ -97,14 +98,18 @@ def receive(frame)
...
+        delta_size = frame[:payload].bytesize
+        delta_size += frame[:padding] || 0
+        @local_window -= delta_size
         emit(:data, frame[:payload]) unless frame[:ignore]
 
+        adjust_recv_window_size(delta_size)
         # Automatically send WINDOW_UPDATE,
         # assuming that emit(:data) can now receive next data
-        window_update(window_size) if window_size > 0
+        if @recv_window_size > 0 && @recv_window_size >= (@local_window / 2)
```
igrigorik: We should document what's happening here, alongside the tradeoffs... E.g we're reducing number of window_update frames, but this does create cases where the server might be blocked because it exhausts the negotiated window and is waiting for us to bump it.

spec/stream_spec.rb
```diff
@@ -598,9 +598,25 @@
       expect(s1.remote_window).to eq 900
     end
 
-    it 'should keep track of incoming flow control' do
+    it 'should not update window when data received is under the threshold' do
```
igrigorik: specify what the threshold is

spec/stream_spec.rb
```diff
@@ -598,9 +598,25 @@
...
+      expect(@client).not_to receive(:send) do |frame|
+        expect(frame[:type]).to eq :window_update
+        expect(frame[:increment]).to eq datalen
+      end
+      @stream.receive HEADERS.deep_dup
+      @stream.receive data
+    end
+
+    it 'should update window when data received is over the threshold' do
+      data1 = DATA.merge(payload: 'a'*16_384, flags: [])
```
igrigorik: Ditto, it's not clear from the test what the threshold is / how this logic is triggered.

lib/http/2/stream.rb
HoneyryderChuck: most of the logic of the code snippet is directly inherited from nghttp, which carries the recv_window_size around. More info about usefulness in the other comments.

HoneyryderChuck: you might be right though, as nghttp doesn't seem to mess with their local_window_size parameters, but [calculate the local window as another operation dependent from the recv_window_size](https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L7113-L7115).

lib/http/2/stream.rb
HoneyryderChuck: This was directly inherited from [here]( https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4904-L4922) and [here](https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4944). 

But maybe, if not counting with recv_window_size, this check is being handled somewhere else?

HoneyryderChuck:
@igrigorik hopefully it looks better now. I'm still uneasy about the `#verify_recv_window_size` method, as I'm mostly following nghttp's logic, but I'm not sure if such a check isn't being performed somewhere else.

lib/http/2/stream.rb
```diff
@@ -97,14 +101,33 @@ def receive(frame)
...
-        @local_window -= window_size
+        # Emit DATA frame
+        delta_size = frame[:payload].bytesize
+        delta_size += frame[:padding] || 0
         emit(:data, frame[:payload]) unless frame[:ignore]
 
-        # Automatically send WINDOW_UPDATE,
-        # assuming that emit(:data) can now receive next data
-        window_update(window_size) if window_size > 0
+        @local_window -= delta_size
```
igrigorik: We should decrement window size before calling emit

lib/http/2/stream.rb
```diff
@@ -86,7 +87,10 @@ def initialize(connection:, id:, weight: 16, dependency: 0, exclusive: false, pa
       @send_buffer = []
 
       on(:window) { |v| @remote_window = v }
-      on(:local_window) { |v| @local_window = v }
+      on(:local_window) do |v|
+        @local_window_size = v
+        @local_window = v
```
igrigorik: @local_window_max? Both are tracking size, the names as is don't make it easy to grok what's going on.

Also, you can simplify this (and above) to: `@local_window_max = @local_window = v`

lib/http/2/stream.rb
```diff
@@ -97,14 +101,33 @@ def receive(frame)
 
       case frame[:type]
       when :data
-        window_size = frame[:payload].bytesize
-        window_size += frame[:padding] || 0
-        @local_window -= window_size
+        # Emit DATA frame
+        delta_size = frame[:payload].bytesize
```
igrigorik: frame_size? why "delta"?

lib/http/2/stream.rb
```diff
@@ -97,14 +101,33 @@ def receive(frame)
...
+        delta_size += frame[:padding] || 0
         emit(:data, frame[:payload]) unless frame[:ignore]
 
-        # Automatically send WINDOW_UPDATE,
-        # assuming that emit(:data) can now receive next data
-        window_update(window_size) if window_size > 0
+        @local_window -= delta_size
+        recv_window_size = @local_window_size - @local_window
```
igrigorik: recv_window_size -> local_window_used, or some such? we're switching nomenclature for no reason.

lib/http/2/stream.rb
```diff
@@ -97,14 +101,33 @@ def receive(frame)
...
+        # verify window size
+        verify_recv_window_size(recv_window_size, delta_size)
```
igrigorik: Let's just inline this, since we're not calling it anywhere else.

lib/http/2/stream.rb
```diff
@@ -587,6 +610,21 @@ def process_priority(frame)
       #   We should add code here.
     end
 
+    # If DATA frame is received with length > 0 and
```
igrigorik: Let's rewrite this with http-2 constants. Also we don't return -1.

spec/stream_spec.rb
```diff
@@ -598,9 +598,25 @@
       expect(s1.remote_window).to eq 900
     end
 
-    it 'should keep track of incoming flow control' do
+    it 'should not update window when data received is half of local window size' do
```
igrigorik: "when received DATA is less than half of maximum local window size"

lib/http/2/stream.rb
HoneyryderChuck: is this `@local_window_max` supposed to be defined? What should I use it for? I thought that was the point of `@local_window_size`, hence the update there.

lib/http/2/stream.rb
HoneyryderChuck: yup, I "inherited" the naming from `nghttp`.

lib/http/2/stream.rb
HoneyryderChuck: again, "inherited" naming from `nghttp`. `local_window_used` is clearer.

HoneyryderChuck:
@igrigorik just had a quick overview on it and did the changes. Naming is hard :) . Still waiting on the point of `@local_window_max`, otherwise I'd say it's ready to go.

igrigorik:
> Still waiting on the point of @local_window_max

Not sure I follow, which part?

lib/http/2/stream.rb
```diff
@@ -77,6 +77,7 @@ def initialize(connection:, id:, weight: 16, dependency: 0, exclusive: false, pa
       @weight = weight
       @dependency = dependency
       process_priority(weight: weight, stream_dependency: dependency, exclusive: exclusive)
+      @local_window_size = connection.local_settings[:settings_initial_window_size]
```
igrigorik: Both @local_window and @local_window_size track "size": one remaining size and other maximum size. Hence my suggestion to rename the latter to @local_window_max. It makes parsing the rest of logic _much_ easier.

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
 
       case frame[:type]
       when :data
+        # Emit DATA frame
         window_size = frame[:payload].bytesize
```
igrigorik: Rename to frame_size, since that's what we're computing here.

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
+        # If DATA frame is received with length > 0 and
+        # current received window size + delta length is strictly larger than
+        # local window size, it throws a stream error for  FLOW_CONTROL_ERROR.
+        # Note that local_window_size is calculated after SETTINGS ACK is
+        # received from peer, so peer must honor this limit. If the resulting
+        # local_window_used is strictly larger than NGHTTP2_MAX_WINDOW_SIZE,
+        # throw a FLOW_CONTROLL_ERROR stream error too.
+        #  (https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4905-L4922)
+        #
+        if local_window_used > @local_window_size - window_size || local_window_used > 0x7fffffff - window_size
```
igrigorik: if local_window_used > @local_window_max - frame_size...

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
         @local_window -= window_size
         emit(:data, frame[:payload]) unless frame[:ignore]
 
-        # Automatically send WINDOW_UPDATE,
-        # assuming that emit(:data) can now receive next data
-        window_update(window_size) if window_size > 0
+        local_window_used = @local_window_size - @local_window
+        # If DATA frame is received with length > 0 and
+        # current received window size + delta length is strictly larger than
+        # local window size, it throws a stream error for  FLOW_CONTROL_ERROR.
```
igrigorik: "..., throw a flow control error." 

You can drop the rest.

lib/http/2/stream.rb
HoneyryderChuck: I'm going to nitpick on top of your nitpick :) and rename it to `local_window_max_size`, as it makes semantically sense (window_max what?).

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
+        # If DATA frame is received with length > 0 and
+        # current received window size + delta length is strictly larger than
+        # local window size, it throws a flow control error.
+        # Note that local_window_max_size is calculated after SETTINGS ACK is
+        # received from peer, so peer must honor this limit. If the resulting
+        # local_window_used is strictly larger than NGHTTP2_MAX_WINDOW_SIZE,
+        # throw a flow control error too.
+        #  (https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4905-L4922)
+        #
+        if local_window_used > @local_window_max_size - frame_size || local_window_used > 0x7fffffff - frame_size
```
igrigorik: Silly question, can we simplify the first check to `@local_window < 0`? :-)
Do we really need the second check? What does it catch that the first one doesn't?

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
+        # (ex: DATA frames with short payloads),
+        # the noise generated by flow control frames creates enough
+        # congestion for this to be deemed very inefficient.
+        #
+        # This heuristic was inherited from nghttp, which delays the
+        # WINDOW_UPDATE until at least half the window is exhausted.
+        # This works because the sender doesn't need those increments
+        # until the receiver window is exhausted, after which he'll be
+        # waiting for the WINDOW_UPDATE frame.
+        if local_window_used > 0 && local_window_used >= (@local_window_max_size / 2)
```
igrigorik: `if @local_window =< (@local_window_max_size / 2)` ? Do we even need local_window_used?

lib/http/2/stream.rb
HoneyryderChuck: where do you see the `@local_window < 0`? Are you sure about that condition?

If you mean the condition for the flow control error, to be frank, I copied it from [here](https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4916-L4917), and if I had to guess, it looks like some spec-compliance check. But I defer to @tatsuhiro-t for the real reason.

lib/http/2/stream.rb
HoneyryderChuck: not in this line at least :)

lib/http/2/stream.rb
igrigorik: There are multiple ways to implement the check; we don't need to verbatim copy each other.

If I understand the intent correctly, the point is to verify that the sender did not exceed the flow control window. Since we decrement frame_size from @local_window.. if the value falls below zero, that's an error.

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
+        # If DATA frame is received with length > 0 and
+        # current received window size + delta length is strictly larger than
+        # local window size, it throws a flow control error.
+        # Note that local_window_max_size is calculated after SETTINGS ACK is
+        # received from peer, so peer must honor this limit. If the resulting
+        # local_window_used is strictly larger than NGHTTP2_MAX_WINDOW_SIZE,
+        # throw a flow control error too.
+        #  (https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4905-L4922)
+        #
+        if @local_window < frame_size || local_window_used > 0x7fffffff - frame_size
```
igrigorik: - First check is not right. You've already decremented frame_size from @local_window. See my earlier comment for checking @local_window > 0 
- What's the point of second clause? What does it cover that first one doesn't?

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
+        # (ex: DATA frames with short payloads),
+        # the noise generated by flow control frames creates enough
+        # congestion for this to be deemed very inefficient.
+        #
+        # This heuristic was inherited from nghttp, which delays the
+        # WINDOW_UPDATE until at least half the window is exhausted.
+        # This works because the sender doesn't need those increments
+        # until the receiver window is exhausted, after which he'll be
+        # waiting for the WINDOW_UPDATE frame.
+        if @local_window < @local_window_max_size && @local_window <= (@local_window_max_size / 2)
```
igrigorik: why would the first clause ever be true? :)

HoneyryderChuck: you mean, not true? yeah, this made more sense in C probably... I don't know if you have to take into account the case when local window is updated by the peer to a value under the current local window. What to do then?

lib/http/2/stream.rb
HoneyryderChuck: fixed. Should I leave the second clause though? I'm not finding a good case where the window used actually surpasses the max window...

lib/http/2/stream.rb
```diff
@@ -97,14 +98,43 @@ def receive(frame)
...
+        # If DATA frame is received with length > 0 and
+        # current received window size + delta length is strictly larger than
+        # local window size, it throws a flow control error.
+        # Note that local_window_max_size is calculated after SETTINGS ACK is
+        # received from peer, so peer must honor this limit. If the resulting
+        # local window used is strictly larger than NGHTTP2_MAX_WINDOW_SIZE,
+        # throw a flow control error too.
+        #  (https://github.com/nghttp2/nghttp2/blob/2bf3680d870953010d7e1e6e4a66510f8458cc3c/lib/nghttp2_session.c#L4905-L4922)
+        #
+        if @local_window < 0 || local_window_used > 0x7fffffff
```
igrigorik: The second check is unnecessary. 
- We will already raise an error if the peer tries to negotiate a window size above max threshold
- The first part of the "or" will catch cases where peer exceeds above limit or the flow control window

Also, please drop all the text starting with "Note that..." — this is nghttp specific.

lib/http/2/stream.rb
igrigorik: Then both are reset. That's exactly what line 90 is doing. 

Let's drop the first clause, afaik it's a noop.

HoneyryderChuck:
@igrigorik fine like this?

HoneyryderChuck:
@igrigorik in a way, it's good that you didn't accept this yet. I'm testing using this at a client, and it hangs when sending multiple requests; the server at some point hangs, waiting on `WINDOW_UPDATE` from the client, and the reason seems to be that we're calculating this only per stream, when the local window size should be the connection window size.

HoneyryderChuck:
@tatsuhiro-t some help? The server in this case is nghttp2.org . This is the relevant part of the framelog:

```
connection (HTTP/2): {:length=>14708, :type=>:data, :flags=>[], :stream=>1, :payload=>14708}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>2789, :type=>:data, :flags=>[], :stream=>3, :payload=>2789}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>5587, :type=>:data, :flags=>[], :stream=>5, :payload=>5587}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>16375, :type=>:data, :flags=>[], :stream=>7, :payload=>16375}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>8385, :type=>:data, :flags=>[], :stream=>9, :payload=>8385}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>16384, :type=>:data, :flags=>[], :stream=>11, :payload=>16384}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>1307, :type=>:data, :flags=>[], :stream=>13, :payload=>1307}
(... hangs ...)
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>19, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>17, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>15, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>1, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>3, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>5, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>7, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>9, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>13, :error=>:internal_error}
connection (HTTP/2): frame was received
connection (HTTP/2): {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>11, :error=>:internal_error}
```

From what I understood, this problem was being hidden before by `http-2` send `WINDOW_UPDATE` per `DATA` frame. When it hangs, the server seems to wait for `WINDOW_UPDATE`, as the total of bytes sent (65535) is the window limit. 

Should the local and remote window be maintained on a connection level, or stream level? It seems that nghttp expects the former.

igrigorik:
@HoneyryderChuck see http://httpwg.org/specs/rfc7540.html#rfc.section.5.2.1

> Flow control is based on WINDOW_UPDATE frames. Receivers advertise how many octets they are prepared to receive on a stream and for the entire connection. This is a credit-based scheme.

HoneyryderChuck:
@igrigorik I'll take that you want to emphasize the "receive on a stream **and** for the entire connection" :) 

What should be the best approach? Should I move window size calculations to the connection?

igrigorik:
Quick sanity check, to make sure I'm following the problem correctly: 

Proposed PR changes window_update behavior to fire only when >1/2 of the stream window is consumed. However, connection window state is currently tracked via per-stream window updates: 

https://github.com/igrigorik/http-2/blob/master/lib/http/2/connection.rb#L667
https://github.com/igrigorik/http-2/blob/master/lib/http/2/connection.rb#L146-L151

As a result, if we have multiple individual streams under the 1/2 threshold, the connection window is exhausted and everything stalls. Correct?

---

This adds more complexity into connection.rb: we'd have to make it similarly aware of sent vs remaining limits (e.g. by doing accounting in `send`), and emit window updates on its own, instead of being tied to stream's window_update calls.

HoneyryderChuck:
> As a result, if we have multiple individual streams under the 1/2 threshold, the connection window is exhausted and everything stalls. Correct?

Correct.

>This adds more complexity into connection.rb ....

Perhaps. But it would do the correct thing, right? It seems that the connection is currently advertising a receive window too large, which just hasn't been a problem yet. 

This does increase the complexity of the changes, though. I thought about doing the accounting [here, where data frames seem to be delegated to the stream](https://github.com/igrigorik/http-2/blob/master/lib/http/2/connection.rb#L312-L315), but now you just mentioned accounting in `send`, and I'm a bit lost here. What changes would you suggest here, so that I don't waste too much time chasing the wrong solution?

igrigorik:
> Perhaps. But it would do the correct thing, right? 

Assuming we want to support the behavior you're after in this PR, yes.

> It seems that the connection is currently advertising a receive window too large, which just hasn't been a problem yet.

That's not the case. Connection flow control worked correctly prior to updates in this PR: every data frame emits a stream window update, and connection window is tracked based on stream updates. The problem now is that we're deferring stream window_update's, which starves the connection window.
 
> This does increase the complexity of the changes, though. I thought about doing the accounting here, where data frames seem to be delegated to the stream, but now you just mentioned accounting in send, and I'm a bit lost here. 

The bit you're looking at is for receiving data, that's not the right place.

> What changes would you suggest here, so that I don't waste too much time chasing the wrong solution?

Let's start by adding a test (or two) to verify that Connection emits a window_update when > 1/2 of the window is used up, independently of the stream updates — i.e reproduce the starvation scenario. From there, we'd probably want to update `send` to perform window accounting, similar to what we have in stream.rb.

HoneyryderChuck:
@igrigorik just added a test to reproduce it, if you can have a look to see whether the case is well covered.

HoneyryderChuck:
Added the logic for the connection-level accounting. Tests are passing, and most important, my local client and server scenarios are also working. I just imported the window accounting logic to the connection when data frames are received, and updated accordingly. 

I think that a few things might be missing, namely that TODO I left there, but also the part where you were mentioning the `#send`. Shouldn't the streams be using the local window of the connection and updating it accordingly, instead of tracking their own?

lib/http/2/connection.rb
```diff
@@ -312,6 +313,17 @@ def receive(data)
           else
             if (stream = @streams[frame[:stream]])
               stream << frame
+              # TODO: when is this NOT a data frame?
```
igrigorik: It can be any stream-level frame, so the check here is correct.

igrigorik:
BTW, can you appease rubucop? :-)

https://travis-ci.org/igrigorik/http-2/jobs/315486702#L1061

HoneyryderChuck:
@igrigorik done. You might have some issues with the method names again, though ;)

> If we normalize @local_window_max_size and @local_window_limit to a single name

I'd prefer to leave this to you, this might have other implications. For now, I'd leave the argument there, and you could set an appropriate name in both resources and remove it before you release a new version, is that ok?

I'll still do a few tests comparing with nghttp, but it should be good now.

igrigorik:
Sorry about the delay. Reviewing the latest update, everything looks great. Thanks for your patience here! :-)

HoneyryderChuck:
No worries, glad to help :)

HoneyryderChuck:
Finally had time to test these changes against the nghttp frame results, and can confirm that it's now generating the same amount of `WINDOW_UPDATE` frames.

igrigorik:
nice! 👍

PR #110, opened 2018-01-08
added check to set path to / if empty

### LINKED ISSUE #106: Client failing to GET https://nghttp2.org
While trying to debug the other issue reported, I've realized that I can't request a publicly accessible website with the simple client (in this case, `https://nghttp2.org`). Just try the example client:

```
>  ruby example/client.rb https://nghttp2.org
Sending HTTP 2.0 request
Sent frame: {:type=>:settings, :stream=>0, :payload=>[[:settings_max_concurrent_streams, 100]]}
Sent frame: {:type=>:headers, :flags=>[:end_headers, :end_stream], :payload=>[[":scheme", "https"], [":method", "GET"], [":authority", "nghttp2.org:443"], [":path", ""], ["accept", "*/*"]], :stream=>1}
[Stream 1]: closing client-end of the stream
Received frame: {:length=>18, :type=>:settings, :flags=>[], :stream=>0, :payload=>[[:settings_max_concurrent_streams, 100], [:settings_initial_window_size, 1048576], [:settings_header_table_size, 8192]]}
Sent frame: {:type=>:settings, :stream=>0, :payload=>[], :flags=>[:ack]}
Received frame: {:length=>0, :type=>:settings, :flags=>[:ack], :stream=>0, :payload=>[]}
Received frame: {:length=>4, :type=>:rst_stream, :flags=>[], :stream=>1, :error=>:protocol_error}
[Stream 1]: stream closed

```

The difference I see to the frame output from `nghttp`is that the latter does the whole "connection dance" before sending the headers frame (send/receive settings and ack, send priorities...).

HoneyryderChuck:
Fixes #106

### DIFF
--- a/lib/http/2/compressor.rb
+++ b/lib/http/2/compressor.rb
@@ -444,7 +444,11 @@ module HTTP2
 
         # Literal header names MUST be translated to lowercase before
         # encoding and transmission.
-        headers.map! { |hk, hv| [hk.downcase, hv] }
+        headers.map! do |hk, hv|
+          key = hk.downcase
+          hv = "/" if key == ":path" && hv.empty?
+          [key, hv]
+        end
 
         commands = @cc.encode(headers)
         commands.each do |cmd|
--- a/lib/http/2/compressor.rb
+++ b/lib/http/2/compressor.rb
@@ -446,7 +446,7 @@ module HTTP2
         # encoding and transmission.
         headers.map! do |hk, hv|
           key = hk.downcase
-          hv = "/" if key == ":path" && hv.empty?
+          hv = '/' if key == ':path' && hv.empty?
           [key, hv]
         end
 
--- a/lib/http/2/compressor.rb
+++ b/lib/http/2/compressor.rb
@@ -199,8 +199,12 @@ module HTTP2
         commands = []
         # Literals commands are marked with :noindex when index is not used
         noindex = [:static, :never].include?(@options[:index])
-        headers.each do |h|
-          cmd = addcmd(h)
+        headers.each do |field, value|
+          # Literal header names MUST be translated to lowercase before
+          # encoding and transmission.
+          field = field.downcase
+          value = '/' if field == ':path' && value.empty?
+          cmd = addcmd(field, value)
           cmd[:type] = :noindex if noindex && cmd[:type] == :incremental
           commands << cmd
           process(cmd)
@@ -220,7 +224,7 @@ module HTTP2
       #
       # @param header [Array] +[name, value]+
       # @return [Hash] command
-      def addcmd(header)
+      def addcmd(*header)
         exact = nil
         name_only = nil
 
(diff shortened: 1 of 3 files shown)


HoneyryderChuck:
@igrigorik tbh I'd prefer to move this check somewhere else, and remove the `#map!` call altogether, as well as every other specific array method call. Ideally, headers could be something responding to `#each` and yielding two parameters, and forcing keys to downcase should either be a concern of the end user, or done at a lower layer, before getting the header key id for encoding.

igrigorik:
Oh, I was actually thinking of adding a condition in the client.rb, not in the library itself. I'm inclined to let the user decide the path for themselves.

> ... forcing keys to downcase should either be a concern of the end user, or done at a lower layer, before getting the header key id for encoding.

Well, where that logic lives today is about as "low level" as it gets, as we're enforcing this as part of the compressor logic.

HoneyryderChuck:
> I'm inclined to let the user decide the path for themselves.

I'd leave it, as the kind of error that the parser triggers is not very obvious from the lib user perspective, and there is (?) no use case covered by the empty ":path".

> Well, where that logic lives today is about as "low level" as it gets, 

I'd argue that [this](https://github.com/igrigorik/http-2/blob/master/lib/http/2/compressor.rb#L203) is the lowest-possible level, which would allow to support "each-able" headers (each block would have to splat field and value though).

lib/http/2/compressor.rb
```diff
@@ -444,7 +444,11 @@ def encode(headers)
 
         # Literal header names MUST be translated to lowercase before
         # encoding and transmission.
-        headers.map! { |hk, hv| [hk.downcase, hv] }
+        headers.map! do |hk, hv|
```
Tr4pSt3R: Hey @HoneyryderChuck is there a corresponding test/spec for this new behaviour please?

HoneyryderChuck: @Tr4pSt3R not until there is agreement that this is a relevant feature.

igrigorik:
> The same principally is used for downcasing the keys, which could also be left to the user but isn't. Both would enforce correctness.

Fair enough. The URI spec [does allow](https://tools.ietf.org/html/rfc3986#section-3.3) for empty path as valid value. The HTTP/2 spec [does hint](http://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.3), although indirectly that path-absolute definition should be used.

> I'd argue that this is the lowest-possible level, which would allow to support "each-able" headers (each block would have to splat field and value though).

To make sure I understand, you're suggesting we move the downcasing and path check into the headers.each look under encode? I guess that should be fine.

HoneyryderChuck:
@igrigorik I went ahead and removed the call to #to_a as well. In this way the argument requirement for `#headers` is just that it responds to `#each`.

HoneyryderChuck:
@igrigorik as @Tr4pSt3R pointed out, there was no test for the new behaviour, is it possible to revert? :)

PR #111, opened 2018-01-09
H2C Upgrade client 

### LINKED ISSUE #99: Feature: Provide helper to build http2-settings header (for h2c)
As of now, it is only possible to encode full frames with the current API:

```ruby
# inside connection
frame = {type: :settings, ...}
encode([frame])....
```

Which is fine for most uses. However, the `h2c` spec allows for encoding client setting in the `HTTP2-settings` frame. This is not the full frame, but only the payload part. This is [a few layers down in the framer](https://github.com/igrigorik/http-2/blob/master/lib/http/2/framer.rb#L223-L235). 

I'd propose to have a public API for returning only the encoded payload of frames (maybe harder), or to add a method to the connection/client to return the payload for the `HTTP2-settings` header. This is my current workaround:

```ruby
# HTTP2 Client extension
def http2_settings
  payload = @local_settings.select { |k, v| v != HTTP2::SPEC_DEFAULT_CONNECTION_SETTINGS[k] }
  frame = { :type => :settings, :stream => 0, payload: Array(payload) }
  encode(frame).map do |f| 
    noheader_frame = f[9..-1] # remove header from http-2 frame
    Base64.urlsafe_encode64(noheader_frame)
  end.join
end	

```

HoneyryderChuck:
Fixes #99 

* Adds `#upgrade` method, which allows client to do the HTTP/2 h2c connection dance and start the first stream in the proper state.
* Adds helper method to generate the proper base64 HTTP2-Settings value

@igrigorik if you agree with the signatures, I'll be happy to provide some tests.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -37,7 +37,7 @@ Metrics/MethodLength:
 # Offense count: 1
 # Configuration parameters: CountKeywordArgs.
 Metrics/ParameterLists:
-  Max: 6
+  Max: 7
 
 # Offense count: 10
 Metrics/PerceivedComplexity:
--- a/example/upgrade_client.rb
+++ b/example/upgrade_client.rb
@@ -1,0 +1,158 @@
+# frozen_string_literals: true
+
+require_relative 'helper'
+require 'http_parser'
+
+options = {}
+OptionParser.new do |opts|
+  opts.banner = 'Usage: upgrade_client.rb [options]'
+end.parse!
+
+uri = URI.parse(ARGV[0] || 'http://localhost:8080/')
+sock = TCPSocket.new(uri.host, uri.port)
+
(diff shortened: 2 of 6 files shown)


igrigorik:
So far so good. I'd love to see tests + example client that exercises this. Also, need to appease rubocop. :)

HoneyryderChuck:
@igrigorik tests, rubocop and example client here! I'd suggest using nghttp2.org, which supports h2c negotiation, for testing purposes.

HoneyryderChuck:
there is a small fix to be made. The `HTTP::Parser` might have buffered HTTP/2 frames in the buffer after reading from the socket, so this data has to be sent to the client connection first thing *after* it is upgrade.

The same has to be done for the server, if it isn't done already.

PR #116, opened 2018-01-22
emitting :promise_headers when promise becomes available

### LINKED ISSUE #114: promise: using the same callback both for request as response headers
I've just noticed that, when trying to read promises from a client, the same callback is used both for when the promise arrives, and when the response headers come:

```ruby
# example/client.rb
conn.on(:promise) do |promise|
  promise.on(:headers) do |h|
    # will get the push headers AND the response headers
    # "promise headers:  [[":method",  "GET"]....
    # "promise headers:  [[":status",  "200"]....
    log.info "promise headers: #{h}"
  end
end
```

I'd propose to solve this by sending the request headers in the same callback, which will be more straightforward:

```ruby
conn.on(:promise) do |promise, request_headers|
  promise.on(:headers) do |response_headers|
...
```

@igrigorik what do you think? is this feasible?

HoneyryderChuck:
Fixes #114

### DIFF
--- a/example/client.rb
+++ b/example/client.rb
@@ -55,6 +55,10 @@ conn.on(:frame_received) do |frame|
 end
 
 conn.on(:promise) do |promise|
+  promise.on(:promise_headers) do |h|
+    log.info "promise request headers: #{h}"
+  end
+
   promise.on(:headers) do |h|
     log.info "promise headers: #{h}"
   end
--- a/lib/http/2/stream.rb
+++ b/lib/http/2/stream.rb
@@ -103,8 +103,10 @@ module HTTP2
         # Emit DATA frame
         emit(:data, frame[:payload]) unless frame[:ignore]
         calculate_window_update(@local_window_max_size)
-      when :headers, :push_promise
+      when :headers
         emit(:headers, frame[:payload]) unless frame[:ignore]
+      when :push_promise
+        emit(:promise_headers, frame[:payload]) unless frame[:ignore]
       when :priority
         process_priority(frame)
       when :window_update
--- a/spec/client_spec.rb
+++ b/spec/client_spec.rb
@@ -86,10 +86,25 @@ RSpec.describe HTTP2::Client do
 
       promise = nil
       @client.on(:promise) { |stream| promise = stream }
-      @client << set_stream_id(f.generate(PUSH_PROMISE.dup), s.id)
+      @client << set_stream_id(f.generate(PUSH_PROMISE.deep_dup), s.id)
 
       expect(promise.id).to eq 2
       expect(promise.state).to eq :reserved_remote
+    end
+
+    it 'should emit promise headers for received PUSH_PROMISE' do
+      header = nil
+      s = @client.new_stream
+      s.send HEADERS.deep_dup
+
+      @client.on(:promise) do |stream|
+        stream.on(:promise_headers) do |h|
+          header = h
+        end
+      end
+      @client << set_stream_id(f.generate(PUSH_PROMISE.deep_dup), s.id)
+
+      expect(header).to eq([%w(a b)])
     end
 
     it 'should auto RST_STREAM promises against locally-RST stream' do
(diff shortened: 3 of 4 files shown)


HoneyryderChuck:
@igrigorik there is something weird going on with the deep_dup implementation for the specs. In some cases, a deep dup'ed copy pased to a method which encodes headers will wipe out the payload from the test frames. I've been experiencing this very randomly with my promise header tests, but have verified that this also happens for the headers frame.

Try running:

```
> bundle exec rspec spec/client_spec.rb:78 spec/stream_spec.rb:28
```

This should fail depending on which spec is executed first. But if you trace the stream spec using byebug, you'll see that the payload is gone:

```ruby
    26:         expect(@stream.state).to eq :open
    27:       end
    28:       it 'should transition to reserved (local) on sent PUSH_PROMISE' do
    29:         @stream.send PUSH_PROMISE.deep_dup
    30:         require "pry-byebug"; binding.pry
 => 31:         expect(@stream.state).to eq :reserved_local
    32:       end
    33:       it 'should transition to reserved (remote) on received PUSH_PROMISE' do
    34:         @stream.receive PUSH_PROMISE
    35:         expect(@stream.state).to eq :reserved_remote
    36:       end

[1] pry(#<RSpec::ExampleGroups::HTTP2Stream::StreamStates::Idle>)> PUSH_PROMISE
=> {:type=>:push_promise, :flags=>[:end_headers], :stream=>1, :promise_stream=>2, :payload=>""}
```

I thought at first that this was from my spec, but the same also happens for `stream_spec.rb:22`:

```ruby
    18: 
    19:     context 'idle' do
    20:       it 'should transition to open on sent HEADERS' do
    21:         @stream.send HEADERS.deep_dup
    22:         require "pry-byebug"; binding.pry
 => 23:         expect(@stream.state).to eq :open
    24:       end
    25:       it 'should transition to open on received HEADERS' do
    26:         @stream.receive HEADERS
    27:         expect(@stream.state).to eq :open
    28:       end

[1] pry(#<RSpec::ExampleGroups::HTTP2Stream::StreamStates::Idle>)> HEADERS
=> {:type=>:headers, :flags=>[:end_headers], :stream=>1, :payload=>""}
```

Thoughts?

HoneyryderChuck:
I find them too frequent and hard to ignore. And as this already happens to the headers, it oughta be rethought a bit. 

The first suggestion would be to move away from using deep-dup'ed hashes, and just go with helper methods which always return a fresh instance of the frame hash. You might have had your reasons to go with deep-dups, so I'd like to hear them before implementing anything.

igrigorik:
> The first suggestion would be to move away from using deep-dup'ed hashes, and just go with helper methods which always return a fresh instance of the frame hash. You might have had your reasons to go with deep-dups, so I'd like to hear them before implementing anything.

Well, to be fair, that's effectively what deep_dup is under the hood, right? I'm open to suggestions for how we can do this better! :)

HoneyryderChuck:
I meant, helper methods which would return fresh instances, instead of relying on deep-dup to work as it should. So, instead of PUSH_PROMISE.deep_due everywhere, one would just have push_promise_frame. Rspec helps inn that by allowing such DSL injection.

What do you think?

HoneyryderChuck:
@igrigorik I gave up rewriting the tests in the PR, as this will open a different can of worms. So let's keep this PR conservative and change the minimum?

I do think that the tests should be rewritten, preferable without deep-dup feature, and using helper methods returning fresh frame instances.

igrigorik:
@HoneyryderChuck current PR seems to pull in some existing changes, could you rebase?

> I do think that the tests should be rewritten, preferable without deep-dup feature, and using helper methods returning fresh frame instances.

That's fair. We can open a separate issue for that.

HoneyryderChuck:
@igrigorik done.

And 👍  for the tests in a separate repo. On the other hand, is that a blocker for a release? (Maybe 0.9.x would make sense due to the changed `:promise_headers` key?)

HoneyryderChuck:
@igrigorik done. Although, there's nothing specific regarding this callback in the README.

igrigorik:
Cool, looks good. Thanks for the great work here!

> And 👍 for the tests in a separate repo. On the other hand, is that a blocker for a release? 

No, this is an existing issue. Let's file a separate thread for it.

> (Maybe 0.9.x would make sense due to the changed :promise_headers key?)

sgtm.

PR #120, opened 2018-02-11
Do not remove closed streams immediately

### LINKED ISSUE #119: Sending RST_STREAM on closed Streams results in GOAWAY
I'm seeing a race condition on pushes of small amounts of data that are canceled by the client. If the push data is small enough and transferred quickly enough, the server may receive a client's RST_STREAM frame after the pushed stream has already been closed. In this case, the http-2 server cannot find the stream id in the `@streams` hash, so it treats it as as unexpected stream identifier and sends GOAWAY with frame 0, killing the connection. Since the closed push stream is actually valid, I believe this is not correct and the RST_STEAM can be silently ignored in this case.

I have a fix for this that will not throw the error if the stream is in the @streams_recently_closed hash.

bengreenberg:
Closed streams need to remain in the Connection @streams list so that
frames received for closed streams do not cause a protocol error.
They are cleaned up if closed for 15 seconds when another stream is
closed. 

Fixes #119

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -656,15 +656,18 @@ module HTTP2
       # permitted to open.
       stream.once(:active) { @active_stream_count += 1 }
       stream.once(:close) do
-        @streams.delete id
         @active_stream_count -= 1
 
         # Store a reference to the closed stream, such that we can respond
         # to any in-flight frames while close is registered on both sides.
         # References to such streams will be purged whenever another stream
         # is closed, with a minimum of 15s RTT time window.
-        @streams_recently_closed.delete_if { |_, v| (Time.now - v) > 15 }
         @streams_recently_closed[id] = Time.now
+        to_delete = @streams_recently_closed.select { |_, v| (Time.now - v) > 15 }
+        to_delete.each do |stream_id|
+          @streams.delete stream_id
+          @streams_recently_closed.delete stream_id
+        end
       end
 
       stream.on(:promise, &method(:promise)) if self.is_a? Server
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -543,6 +543,20 @@ RSpec.describe HTTP2::Connection do
       end.to raise_error(ProtocolError)
     end
 
+    it 'should not raise an error on frame for a closed stream ID' do
+      srv = Server.new
+      srv << CONNECTION_PREFACE_MAGIC
+
+      stream = srv.new_stream
+      stream.send HEADERS.dup
+      stream.send DATA.dup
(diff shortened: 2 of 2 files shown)


igrigorik:
> I'm seeing a race condition on pushes of small amounts of data that are canceled by the client. If the push data is small enough and transferred quickly enough, the server may receive a client's RST_STREAM frame after the pushed stream has already been closed. In this case, the http-2 server cannot find the stream id in the @streams hash, so it treats it as as unexpected stream identifier and sends GOAWAY with frame 0, killing the connection. Since the closed push stream is actually valid, I believe this is not correct and the RST_STEAM can be silently ignored in this case.

Makes sense.

> Looking into this more, I believe this was introduced with #88 . I don't think the intent of that patch was the delete closed immediately. My fix only deletes the stream if its been closed for 15 seconds or longer.

We do want to remove it from the `@streams` list of active streams, to keep the accounting correct. We should use existing `@streams_recently_closed` for what you're after.

We already handle a similar case for push_promises:

https://github.com/igrigorik/http-2/blob/0aeaac237a7dd589aacdee7756759ceb20c378f8/lib/http/2/connection.rb#L273-L290

To cover the case that you're after, we probably need to look at..

https://github.com/igrigorik/http-2/blob/0aeaac237a7dd589aacdee7756759ceb20c378f8/lib/http/2/connection.rb#L331-L333

bengreenberg:
I believe stream accounting is handled by:

```
      stream.once(:close) do
        @active_stream_count -= 1
```

Which i kept in with this patch. Does that handle it?

My original patch actually added an if statement around that `connection_error` line. Unfortunately, that triggered Rubocop to complain about too many layers of nesting, and I think resolving that would trigger a major code refactor. While exploring other options, I noticed that there is handling in the Stream class for streams in a closed state, so I thought it would be good for frames to still make their to the actual Stream. Otherwise, that can be removed because I believe it will never be triggered.

Let me know if you think it's still best to do the `@streams_recently_closed` check before the error is thrown and I will make those changes -- though the refactor is a bit intimidating as I'm new to the codebase.

igrigorik:
> My original patch actually added an if statement around that connection_error line. Unfortunately, that triggered Rubocop to complain about too many layers of nesting, and I think resolving that would require a major code refactor.

Rubocop is directional advice, we can tell it to shush.. :) 

> While exploring other options, I noticed that there is handling in the Stream class for streams in a closed state, so I thought it would be good for frames to still make their way to the actual Stream. Otherwise, I think that code can be removed because I believe it will never be triggered.

Ah, good point. Yes that makes sense.

On second pass, I think this looks reasonable 👍

PR #121, opened 2018-03-17
Fix warnings in lib.

ioquatix:
You might like to take a closer look at the `alias error ...` case. I removed the definition that was being overwritten, but that might not be what you were expecting.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -2,3 +2,4 @@ autotest
 --color
 --format documentation
 --order rand
+--warnings
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -49,9 +49,6 @@ module HTTP2
     # Connection state (:new, :closed).
     attr_reader :state
 
-    # Last connection error if connection is aborted.
-    attr_reader :error
-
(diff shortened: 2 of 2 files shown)


ioquatix:
@igrigorik this could also use some attention.

igrigorik:
@ioquatix apologies about the delay. What are the warnings being fixed here?

ioquatix:
Well, if you add --warnings to rspec, you can see them for yourself. IF you review the PR, you can see there is the same method defined twice with different implementations, and a variable which is not initialised correctly.

igrigorik:
@ioquatix ahh.. missed the conflict with later alias. Interesting, this means that there isn't a direct way to access the last error.. but if nobody complained about that, I guess it's safe to remove. LGTM.

ioquatix:
You could rename it to be ‘last_error’?

igrigorik:
True. Let's see if anyone complains first.. :)

HoneyryderChuck:
My 2 cents: the close callback already yields the last error when available. As it is, the last error getter is of little value to me.

ioquatix:
Okay, all good points 👍

Issue #122, opened 2018-03-20
Failing client requests


ioquatix:
I'm using the latest published gem (0.9.0).

I'm working on an asynchronous implementation which has feature parity with HTTP1 (connection per request and sequential connections using keep-alive). I'm working on the HTTP2 implementation which for the most part appears to work except the functionality depends on when I start the read loop.

Here are two dumps from my client:

```
Sent frame: {:type=>:settings, :stream=>0, :payload=>[[:settings_max_concurrent_streams, 100]]}
Sent frame: {:type=>:headers, :flags=>[:end_headers, :end_stream], :payload=>{":method"=>"GET", ":path"=>"/", ":scheme"=>"https", ":authority"=>"www.codeotaku.com", "accept"=>"*/*", "user-agent"=>"nghttp2/1.30.0"}, :stream=>1}

-- start read loop here

Received frame: {:length=>18, :type=>:settings, :flags=>[], :stream=>0, :payload=>[[:settings_max_concurrent_streams, 128], [:settings_initial_window_size, 65536], [:settings_max_frame_size, 16777215]]}
Sent frame: {:type=>:settings, :stream=>0, :payload=>[], :flags=>[:ack]}
Received frame: {:length=>4, :type=>:window_update, :flags=>[], :stream=>0, :increment=>2147418112}
Received frame: {:length=>0, :type=>:settings, :flags=>[:ack], :stream=>0, :payload=>[]}

[...]

PR #126, opened 2018-04-12
Fix client when it initially receives data rather than sending it

ioquatix:
The connection preface wasn't sent if the first operation was to receive data. This is due to a state machine transition to `:connected` in the frame receive mechanism without invoking `#send_connection_preface`.

As an aside, I sort of think it's a bit inefficient to invoke this function on every data receive/send, but perhaps that's an issue for another time.

### DIFF
--- a/lib/http/2/client.rb
+++ b/lib/http/2/client.rb
@@ -38,7 +38,7 @@ module HTTP2
       super
     end
     
-    def <<(data)
+    def connection_management(frame)
       send_connection_preface
       super
     end
--- a/lib/http/2/client.rb
+++ b/lib/http/2/client.rb
@@ -35,7 +35,12 @@ module HTTP2
     # @param frame [Hash]
     def send(frame)
       send_connection_preface
-      super(frame)
+      super
+    end
+    
+    def <<(data)
+      send_connection_preface
+      super
     end
 
     # sends the preface and initializes the first stream in half-closed state
--- a/lib/http/2/client.rb
+++ b/lib/http/2/client.rb
@@ -39,7 +39,7 @@ module HTTP2
       send_connection_preface
       super
     end
-    
+
     def connection_management(frame)
       send_connection_preface
       super
--- a/lib/http/2/client.rb
+++ b/lib/http/2/client.rb
@@ -50,6 +50,13 @@ module HTTP2
       new_stream(state: :half_closed_local)
     end
 
+    def self.settings_header(**settings)
+      frame = Framer.new.generate(type: :settings, stream: 0, payload: settings)
+      Base64.urlsafe_encode64(frame[9..-1])
+    end
+
+    private
+
     # Emit the connection preface if not yet
     def send_connection_preface
       return unless @state == :waiting_connection_preface
@@ -58,11 +65,6 @@ module HTTP2
 
       payload = @local_settings.select { |k, v| v != SPEC_DEFAULT_CONNECTION_SETTINGS[k] }
       settings(payload)
-    end
-
-    def self.settings_header(**settings)
-      frame = Framer.new.generate(type: :settings, stream: 0, payload: settings)
-      Base64.urlsafe_encode64(frame[9..-1])
     end
   end
 end
--- a/lib/http/2/client.rb
+++ b/lib/http/2/client.rb
@@ -25,6 +25,8 @@ module HTTP2
       @local_role   = :client
       @remote_role  = :server
 
+      @connection_preface_sent = false
+
       super
     end
 
@@ -59,8 +61,10 @@ module HTTP2
 
     # Emit the connection preface if not yet
     def send_connection_preface
-      return unless @state == :waiting_connection_preface
-      @state = :connected
+      return if @connection_preface_sent
+
+      @connection_preface_sent = true
+
       emit(:frame, CONNECTION_PREFACE_MAGIC)
 
       payload = @local_settings.select { |k, v| v != SPEC_DEFAULT_CONNECTION_SETTINGS[k] }
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -350,8 +350,9 @@ module HTTP2
         end
       end
 
+    rescue Error::Error
+      raise
     rescue => e
-      raise if e.is_a?(Error::Error)
       connection_error(e: e)
     end
     alias << receive
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -8,12 +8,14 @@ RSpec.describe HTTP2::Connection do
   let(:f) { Framer.new }
 
   context 'initialization and settings' do
-    it 'should raise error if first frame is not SETTINGS' do
-      (FRAME_TYPES - [SETTINGS]).each do |frame|
+    (FRAME_TYPES - [SETTINGS]).each do |frame|
+      it "should raise error if first frame is #{frame[:type]}" do
         frame = set_stream_id(f.generate(frame.deep_dup), 0x0)
         expect { @conn.dup << frame }.to raise_error(ProtocolError)
       end
-
+    end
+    
+    it 'should not raise error if first frame is SETTINGS' do
       expect { @conn << f.generate(SETTINGS.dup) }.to_not raise_error
       expect(@conn.state).to eq :connected
     end
@@ -50,13 +52,18 @@ RSpec.describe HTTP2::Connection do
       settings = SETTINGS.dup
       settings[:payload] = [[:settings_header_table_size, 256]]
 
-      expect(@conn).to receive(:send) do |frame|
-        expect(frame[:type]).to eq :settings
-        expect(frame[:flags]).to eq [:ack]
-        expect(frame[:payload]).to eq []
+      # We expect two frames here - one for the connection setup, and one for the settings ack.
+      frames = []
+      expect(@conn).to receive(:send).twice do |frame|
+        frames << frame
       end
 
       @conn << f.generate(settings)
+
+      frame = frames.last
+      expect(frame[:type]).to eq :settings
+      expect(frame[:flags]).to eq [:ack]
+      expect(frame[:payload]).to eq []
     end
   end
 
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe HTTP2::Connection do
         expect { @conn.dup << frame }.to raise_error(ProtocolError)
       end
     end
-    
+
     it 'should not raise error if first frame is SETTINGS' do
       expect { @conn << f.generate(SETTINGS.dup) }.to_not raise_error
       expect(@conn.state).to eq :connected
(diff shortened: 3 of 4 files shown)


ioquatix:
An alternative approach is presented in https://github.com/igrigorik/http-2/pull/44/commits/02ba3f089363c2869d211faa5e24bb7f286d2f13

Either approach is fine. However once this PR is merged, the other can be closed. I don't know a lot about the internal details of how `http-2` works, but it appears that `connection_management` could be a more appropriate place for this logic?

ioquatix:
I've added a final commit which implements the alternative (better?) logic. Discussion is required.

ioquatix:
It looks like there are some logic errors - @igrigorik @HoneyryderChuck can you check it and let me know?

ioquatix:
Okay, I've got all the specs passing with minimal changes. It would be good to discuss this further though, there are some issues I think need further discussion:

- The `connection_spec.rb` really seems to test the `class Client`.
- The `Connection` base class is broken in the sense that it doesn't define `@state`.
- There are some "odd" code paths in `Connection`, e.g. `self.is_a? Server` for tracking if stream IDs are valid or not.

ioquatix:
Stuff like

```ruby
RSpec.describe HTTP2::Connection do
  before(:each) do
    @conn = Client.new
  end

  it '...' do
    @conn.send ...
  end
```

sort of seem a bit crappy. It would make more sense to be like:

```ruby
RSpec.describe HTTP2::Client do
  it '...' do
    subject.send ...
  end
end
```

spec/connection_spec.rb
```diff
@@ -50,13 +52,18 @@
...
-        expect(frame[:payload]).to eq []
+      # We expect two frames here - one for the connection setup, and one for the settings ack.
+      frames = []
+      expect(@conn).to receive(:send).twice do |frame|
+        frames << frame
       end
 
       @conn << f.generate(settings)
+
+      frame = frames.last
```
igrigorik: We should also check that first frame is the preface?

lib/http/2/client.rb
```diff
@@ -45,19 +52,23 @@ def upgrade
...
+      frame = Framer.new.generate(type: :settings, stream: 0, payload: settings)
+      Base64.urlsafe_encode64(frame[9..-1])
+    end
+
+    private
+
     # Emit the connection preface if not yet
     def send_connection_preface
-      return unless @state == :waiting_connection_preface
-      @state = :connected
```
igrigorik: Why did you remove the `@state` transition?

lib/http/2/connection.rb
```diff
@@ -350,8 +350,9 @@ def receive(data)
         end
       end
 
+    rescue Error::Error
+      raise
     rescue => e
-      raise if e.is_a?(Error::Error)
```
igrigorik: ?

igrigorik:
FWIW, the logic in https://github.com/igrigorik/http-2/commit/02ba3f089363c2869d211faa5e24bb7f286d2f13 seems simpler. Avoiding the need for new instance variables seems like a plus. Also, squinting, seems like perhaps this should all live in connection instead? WDYT?

https://github.com/igrigorik/http-2/blob/2c127aec0cff711b7f27e9a7f8dad2d952984fd7/lib/http/2/connection.rb#L367

lib/http/2/connection.rb
ioquatix: Exceptions already match based on class, so it seems like a code smell to do this explicitly.

lib/http/2/client.rb
ioquatix: Because it broke too many specs and it seems like something which doesn't really relate directly to state, but is more like, have we sent it yet? Feel free to dig into it and propose a different solution.

spec/connection_spec.rb
ioquatix: Yeah, why not? It's good.

ioquatix:
I'd really need to spend a few days digesting how the entire thing works in more detail in order to really offer an opinion. I just made the change that I thought made the most sense and worked through all the broken specs.

You've (ideally) got a better understanding of the design of the library so I guess if you want a deeper discussion, I'm not the right person... at least, not unless I spend a few days digging into it.

I actually decided to resolve this issue in async-http by not calling read until something was written. I'm not sure it's ideal, but it's one option which avoids this whole state problem.

Honestly, I think there are some issues w.r.t. how Connection/Server/Client work together - that I mentioned earlier. It might be worth reviewing exactly how you want that to work - e.g. is `@state` really necessary?

Additionally, today I ran into race conditions within the library, if you don't register event handlers before calling read, you end up with race conditions, e.g. it's possible that you loose incoming requests, because you didn't call `on(:stream) ...` before reading incoming data.

It's quite tricky to get right and not very obvious. I'm personally a fan that once you construct an object, it's ready to go. I don't like (registering) event handlers (very much).

ioquatix:
Thanks for your feedback.

> I think the PR has a lot of moving parts (code reorder, new ivars...) that don't have to do with it's intended bugfix

There's only one unrelated bit which is the error handling. Basically, everything else is related to the bug fix.

>  implied what could be the fix for this particular issue, which you have implemented in your code with success, so the PR should contain that, and the necessary fixes for the tests to work. 

That's almost exactly what this is. The final approach I took was the one suggested in https://github.com/igrigorik/http-2/commit/02ba3f089363c2869d211faa5e24bb7f286d2f13

Everything else is essentially fixing problems either with the specs, or issues introduced by the change. There are a few minor changes to the specs to make it clearer which ones are failing - in the case we had one spec for about 10 different scenarios, I turned that into a loop so we have 10 different specs, and that's because about half of them were failing and the other half were okay.

> That's not a race condition, it's how the lib works. You do have to set all the callbacks before starting to decode frames.

It's absolutely a race condition, but it's also how the lib works :) The order of (non-deterministic) operations changes the results. In `async-http` in about 90% of cases the server would pick up the request just fine, but if the data was already in the socket when the server started reading, the function call to `on(:stream)` would not be registered in time. Anyway, I've worked around it in my code.

> That's how "user code" interfaces with "lib code", at least for now. However, an alternative would be to pass an object with the callbacks already set up as methods, the way that the http_parser.rb gem does (cc @igrigorik ). However, again, this should be addressed in a separate issue/PR.

Yeah, I agree, it's another possible option. However, at this point, I'm not sure it's worth changing the entire API just because it might be more ergonomic. I think the best option would be to break it out into separate gems, e.g. the framer, hpack, and so on, so the high level implementation can be handled by the client code (with some defaults like `Client` and `Server` that basically work out of the box). Anyway, this discussion is for another time :) Let's focus on whether we can fix this bug in `Client`.

ioquatix:
@igrigorik 

> FWIW, the logic in 02ba3f0 seems simpler. Avoiding the need for new instance variables seems like a plus. Also, squinting, seems like perhaps this should all live in connection instead? WDYT?

Yes, that's what I finally implemented.

However, that change introduces many issues with specs because changing `@state` caused odd issues in `Connection` which I wasn't really able to make sense of. I tried using an explicit variable to track whether it's sent (IMHO a better approach since it's more isolated) and it worked better and caused less issues with specs. I think if you want to try either option and see for yourself what specs break, perhaps you have a better idea of how to fix the state machine. Personally, I find the state machine pretty tricky to understand without really spending a lot of time working through it.

HoneyryderChuck:
> There's only one unrelated bit which is the error handling...

Have only taken an overlook, but I see a method in client.rb being moved, some `rescue` clause changing, and a bit of formatting in the specs, that don't seem to have to do with the contextual change. Again, not saying that it's irrelevant, but I'd say it could be addressed separately. 

> The final approach I took was the one suggested in 02ba3f0

Fair enough. I'd say that my suggestion could take less LOCs :) But I'm ok with whatever works. 

> The order of (non-deterministic) operations changes the results...

I didn't get this part. What is exactly non-deterministic about it? Unless you're handling streams in multiple threads, I'd say this shouldn't be an issue.

ioquatix:
> I didn't get this part. What is exactly non-deterministic about it? Unless you're handling streams in multiple threads, I'd say this shouldn't be an issue.

- Call read - socket buffer empty - no streams started
- Set stream handler
- Call read - socket has data - stream started

vs

- Call read - socket buffer has data - stream missed
- Set stream handler
- Call read - socket has no data - stream not started

> Have only taken an overlook, but I see a method in client.rb being moved, 

You mean, I made it private? As you suggested here? https://github.com/igrigorik/http-2/issues/122#issuecomment-376563850

> some rescue clause changing

Yes, this could be separate.

> and a bit of formatting in the specs, that don't seem to have to do with the contextual change.

This was a useful change to help figure out what parts of the initial change were causing breakage.

HoneyryderChuck:
> You mean, I made it private? As you suggested here? 

My bad, I got confused while reading the diff.

Another thing that I left out before: I don't think you need to use `@connection_preface_sent`, if you either follow my suggestion, or follow [a previously linked suggestion](https://github.com/igrigorik/http-2/commit/02ba3f089363c2869d211faa5e24bb7f286d2f13).

Also, I don't think there is non-determinism there. If you start reading from a socket, you better have a fully-configured parser before you pass it the bytes; specifically, you don't "miss-streams-then-set-up-handler", handler has to be set-up once you start reading from the socket.

ioquatix:
> Another thing that I left out before: I don't think you need to use @connection_preface_sent, if you either follow my suggestion, or follow a previously linked suggestion.

Yeah but I tried that and it and more specs failed. Is it worth trying to figure out why it's not working?

> Also, I don't think there is non-determinism there. If you start reading from a socket, you better have a fully-configured parser before you pass it the bytes; specifically, you don't "miss-streams-then-set-up-handler", handler has to be set-up once you start reading from the socket.

Fair enough. If I was implementing this it would buffer the events until they were dequeued, but I'd also use a different event handling mechanism (probably a queue).

HoneyryderChuck:
> Yeah but I tried that and it and more specs failed. Is it worth trying to figure out why it's not working?

I think so. If you recall from my comments in the linked issue, I did say that the fix would be easy, but the tests part would be the worse :) I'll see if I can reproduce this locally to see exactly how big the gap is.

>  If I was implementing this it would buffer the events until they were dequeued,

The lib already does this, just that "dequeuing" is in this case "emitting", and you need behaviour set for this.

HoneyryderChuck:
I've just run a clean master with my suggested diff:

```ruby
# client.rb
+    def receive(*)
+      send_connection_preface
+      super
+    end
+
```

surprisingly, and against my expectation, it "just worked" (tests passed).

ioquatix:
That's awesome, I wonder why it didn't work for me. I will have to check the original commits.

ioquatix:
Hmm, I added it to `<<`. I guess that's different.

HoneyryderChuck:
ouch, I see what happened. Will come back to it.

ioquatix:
I didn't know `alias` worked like that to be honest. But, it sort of makes sense, it becomes `define_method(new_name, self.instance_method(original_name))` or something to that effect.

igrigorik:
@HoneyryderChuck @ioquatix thanks for digging into this. Been offline for a while, catching up.. what are the next steps here?

ioquatix:
I think this PR takes things in the right direction, but whether you try to make `@state` work or just use a separate variable is still in question. Personally, I think `@state` is the wrong place to manage this transition.

ioquatix:
Once https://github.com/igrigorik/http-2/pull/131 is merged, I will rework this PR.

PR #133, opened 2018-07-02
Upgraded to Rubocop 0.57.2 (no offenses detected).

byronformwalt:
@igrigorik Here are the Rubocop updates provided in a separate PR for your review, as requested.

### DIFF
--- a/example/upgrade_server.rb
+++ b/example/upgrade_server.rb
@@ -39,12 +39,12 @@ end
 
 class UpgradeHandler
   VALID_UPGRADE_METHODS = %w(GET OPTIONS).freeze
-  UPGRADE_RESPONSE = <<-RESP
-HTTP/1.1 101 Switching Protocols
-Connection: Upgrade
-Upgrade: h2c
+  UPGRADE_RESPONSE = <<-RESP.strip_heredoc.freeze
+    HTTP/1.1 101 Switching Protocols
+    Connection: Upgrade
+    Upgrade: h2c
 
-RESP
+  RESP
 
   attr_reader :complete, :headers, :body, :parsing
 
@@ -191,7 +191,7 @@ loop do
         conn << data
       end
 
-    rescue => e
+    rescue StandardError => e
       puts "Exception: #{e}, #{e.message} - closing socket."
       puts e.backtrace.last(10).join("\n")
       sock.close
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -20,9 +20,9 @@ module HTTP2
 
   DEFAULT_CONNECTION_SETTINGS = {
     settings_header_table_size:       4096,
-    settings_enable_push:             1,     # enabled for servers
+    settings_enable_push:             1,                     # enabled for servers
     settings_max_concurrent_streams:  100,
-    settings_initial_window_size:     65_535, #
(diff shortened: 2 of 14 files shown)


example/upgrade_server.rb
```diff
@@ -39,12 +39,12 @@ def request_header_hash
 
 class UpgradeHandler
   VALID_UPGRADE_METHODS = %w(GET OPTIONS).freeze
-  UPGRADE_RESPONSE = <<-RESP
-HTTP/1.1 101 Switching Protocols
-Connection: Upgrade
-Upgrade: h2c
+  UPGRADE_RESPONSE = <<-RESP.strip_heredoc.freeze
```
ioquatix: It might be feasible to use `<<~RESP` which achieves the same thing no? What version of Ruby does it go back to? (and does it really matter for this example client/server)

lib/http/2/connection.rb
```diff
@@ -20,9 +20,9 @@ module HTTP2
 
   DEFAULT_CONNECTION_SETTINGS = {
     settings_header_table_size:       4096,
-    settings_enable_push:             1,     # enabled for servers
+    settings_enable_push:             1,                     # enabled for servers
     settings_max_concurrent_streams:  100,
-    settings_initial_window_size:     65_535, #
```
ioquatix: Mysterious invisible comment :)

lib/http/2/connection.rb
```diff
@@ -496,7 +496,7 @@ def validate_settings(role, settings)
           # allowed frame size (2^24-1 or 16,777,215 octets), inclusive.
           # Values outside this range MUST be treated as a connection error
           # (Section 5.4.1) of type PROTOCOL_ERROR.
-          unless 16_384 <= v && v <= 16_777_215
+          unless v >= 16_384 && v <= 16_777_215
```
ioquatix: Maybe using a range here would be easier to read?

ioquatix:
@igrigorik Should we be concerned that JRuby is failing?

ioquatix:
Okay @igrigorik here is the reason why https://github.com/jruby/jruby/issues/5236

example/upgrade_server.rb
igrigorik: I believe <<~ was introduced in 2.3. Technically, 2.1 and 2.2 are already EOL, so we _could_ require 2.3+ in the gem. That said, is it worth it?

Is Rubocop complaining about this one?

lib/http/2/connection.rb
igrigorik: Yeah, that would probably make things easier to grok. 

Do you know if there is any noticeable perf difference? Microbenchmark to the rescue?

igrigorik:
> Okay @igrigorik here is the reason why jruby/jruby#5236

Thanks for tracking that down. Hopefully it'll be back to normal soon.

example/upgrade_server.rb
igrigorik: Given that [2.2 is EOL](https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/). I'm OK with locking http-2 to 2.2+. If we go down this path, we should update the travis config and gemspec files as well.

Any objections? :)

Issue #129, opened 2018-05-03
goaway when talking to Cloudflare


ioquatix:
When I tried to send the headers frame twice, I received `goaway`.

I ended up doing this:

```
			class Merged
				def initialize(*all)
					@all = all
				end
				
				def each(&block)
					@all.each do |headers|
						headers.each do |key, value|
							yield key, value.to_s
						end
					end
				end
			end
```

However, is there some reason why sending headers twice would be a problem? After my discussion with @HoneyryderChuck I thought it was ok (with `end_headers:false`).

HoneyryderChuck:
Could you print a "frame trace"? That's interesting indeed.

ioquatix:
Let me see if I can repro... I sort of trashed all that code now.

ioquatix:
Okay first problem. If headers is an empty hash, it fails:

```
     1.1) Failure/Error: stream.headers(headers.all[1], end_stream: true)
          
          NoMethodError:
            undefined method `force_encoding' for {}:Hash
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/buffer.rb:70:in `block (2 levels) in <class:Buffer>'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/framer.rb:203:in `generate'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/connection.rb:398:in `block in encode'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/connection.rb:398:in `map'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/connection.rb:398:in `encode'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/connection.rb:381:in `send'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/client.rb:38:in `send'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/emitter.rb:34:in `block in emit'
          # /Users/samuel/.rvm/gems/ruby-2.5.0/gems/http-2-0.9.0/lib/http/2/emitter.rb:33:in `delete_if'
[...]

Issue #128, opened 2018-05-02
How to deal with idempotent requests which fail?


ioquatix:
I'm implementing an h2 proxy.

I'm following the rough outline here: https://tools.ietf.org/html/rfc7230#section-6.3.1

What's interesting to me is how to handle persistent connections which have been closed.

With HTTP/1.x, if you try to write the request line, but it fails, you generally know that the persistent connection was closed - it's not an error, and it's safe to open a new connection and resend the request (even non-idempotent ones).

It would be interesting to know how to handle this with this gem.

At what point can we assume that the request has been received by the remote end?

I assume that after calling:

```ruby
stream.headers(headers, end_stream: true)
io.flush
```

We have to assume at this point the remote end has received the request and we can no longer retry non-idempotent requests.

Thoughts?

igrigorik:
> With HTTP/1.x, if you try to write the request line, but it fails, you generally know that the persistent connection was closed

What does "fails" means in this scenario?

FWIW, H2 introduced GOAWAY to address this case: https://httpwg.org/specs/rfc7540.html#rfc.section.6.8

> Endpoints SHOULD always send a GOAWAY frame before closing a connection so that the remote peer can know whether a stream has been partially processed or not. For example, if an HTTP client sends a POST at the same time that a server closes a connection, the client cannot know if the server started to process that POST request if the server does not send a GOAWAY frame to indicate what streams it might have acted on.

igrigorik:
If the other end closes the underling TCP connection properly you'll get a TCP RST, which can then signal the teardown up the stack. On the other hand, if the connection drops without a RST there isn't much you can do there short of listening for the TCP timeout.. Of course, none of this is h2 specific.

igrigorik:
@ioquatix fyi, might be of interest: https://tools.ietf.org/html/draft-nottingham-httpbis-retry-00

ioquatix:
Thanks that's really interesting.

I've got something that works pretty well right now. I may revisit for the future and that guide will surely be helpful.

ioquatix:
You might appreciate https://news.ycombinator.com/item?id=16964907

igrigorik:
^ that's a good one 😆

PR #134, opened 2018-07-04
Add strict mode flag.

ioquatix:
Add a `strict:` flag to `Connection`. We can then make an efficiency/conformance trade off according to users requirements, and allow users to test non-conformance (by setting strict: false so that invalid behaviour would not be rejected).

I suggest we use this flag initially to gate the following checks:

- Ensure correct order of headers on both client request/response and server request/response (whether this sorts input data or just fails on invalid input still needs to be figured out).
- Ensure that streams don't enter invalid states or make invalid transitions (e.g. sending headers twice).
- Any other high level RFC conformance which can be done in a reasonably efficient way.

This PR simply adds the flag, and it defaults to false. Depending on how the rest of the checks are implemented, we could make the default true.

If we can get consensus on this approach, I will merge it and then we can work on better RFC compliance. I would be happy to hear any arguments for making the default `true` or `false`. I initially thought `true` would be good, but I don't want to risk breaking upstream code. If it's opt in, at least upstream users are making a conscious decision that it's what they want. Maybe in the future it could be the default.

cc @HoneyryderChuck @byronformwalt @ostinelli @kenichi @igrigorik

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -46,6 +46,9 @@ module HTTP2
     include Emitter
     include Error
 
+    # Enforce RFC conformant behaviour.
+    attr_reader :strict
+
     # Connection state (:new, :closed).
     attr_reader :state
 
@@ -69,7 +72,9 @@ module HTTP2
 
     # Initializes new connection object.
     #
-    def initialize(**settings)
+    def initialize(strict: false, **settings)
+      @strict = strict
+
       @local_settings  = DEFAULT_CONNECTION_SETTINGS.merge(settings)
       @remote_settings = SPEC_DEFAULT_CONNECTION_SETTINGS.dup
 


ioquatix:
> @igrigorik what is the expected role and behavior of http-2 against normative requirements of HTTP/2 protocol?

It should conform to the spec where it is both correct and efficient to do so, within the expectations of how this library should be used.

Conformance with the spec requires multiple layers to work together. There are definitely issues where this library, despite how hard you (and others) have worked (and I appreciate it) has conformance issues:

- https://github.com/igrigorik/http-2/issues/129
- https://github.com/igrigorik/http-2/issues/128
- https://github.com/igrigorik/http-2/issues/122

Coming back to your `#downcase` example. If you document in the function "All headers should be lower case strings", is that good enough? In theory it should be, because if users violate that contract then they risk the remote end rejecting the request. `#downcase` is not free, performance wise. If users are already supplying lower case strings, it's superfluous.

That's why I think `strict:` is a good compromise. For the majority of users, you can enable it and it will work in a way which is less efficient but guarantees conformance.

The result would be something like this:

```
field = field.downcase if @strict
```

or something like:

```
pseudo, others = headers.partition{|k,v| k.start_with? ':'} if @strict
```

For higher level libraries who can follow the invariants specified by the functions required for conformance, we will gain performance advantage, or if you prefer to test failures, you can generate invalid requests to check the result.

> Extending the same logic, I would propose that we apply the same logic to header sorting. There is only one h2 compliant ordering of headers, sorting is similarly 'safe' and does not affect semantics of the request, and we can enforce this invariant in http-2 — there's no reason to ask every upstream library to implement same logic.

You are not totally correct on this point. If the sort is unstable (which is why if anything should use `#partition`), it might change the relative order of headers with the same key which CAN change the semantics of a request, which is why hash tables are generally a terrible idea for headers.

>    A recipient MAY combine multiple header fields with the same field
>    name into one "field-name: field-value" pair, without changing the
>    semantics of the message, by appending each subsequent field value to
>    the combined field value in order, separated by a comma.  The order
>    in which header fields with the same field name are received is
>    therefore significant to the interpretation of the combined field
>    value; a proxy MUST NOT change the order of these field values when
>    forwarding a message. (RFC 7230)

An example of where this gets tricky would be the following request:

```
{'key' => 1, 'Key' => 2}.sort.collect{|k,v| [k.downcase, v]}
=> [["key", 2], ["key", 1]]
```

If not carefully thought out, such interactions yield incorrect behaviour (headers in wrong order). Of course, this is solved by using `#partition` instead, it was just an example to show how even the best intentions can go horribly wrong as we start to layer more complex functionality into the library.

ioquatix:
> there's no reason to ask every upstream library to implement same logic.

The upstream libraries I've checked, even non-Ruby ones, simply construct their requests using an Array of key/value pairs which conforms to the order requirements of the RFC, so I'm not sure what logic you are thinking every upstream library is having to deal with? I guess we could do something like:

```
if headers.is_a? Hash
    headers = headers.sort
end
```

PR #140, opened 2018-08-22
ignore management frames for a period after closing a connection

### LINKED ISSUE #139: receiving ping while closing connection causing ProtocolError
I've ran into the case where my connection receives a ping frame just after I call `goaway` on it. More or less this sequence:

```
1: {:length=>0, :type=>:data, :flags=>[:end_stream], :stream=>1, :payload=>0}
1: <- DATA: 0 bytes...
1: <- ""
# here I start cleaning up, as I won't need the connection anymore
: frame was buffered!
: {:type=>:goaway, :last_stream=>0, :error=>:no_error, :payload=>nil}
1: closing stream
READ: 17 bytes...
0: frame was received!
0: {:length=>8, :type=>:ping, :flags=>[], :stream=>0, :payload=>"\x00\x00\x00\x00\x00\x00\x00\x00"}
ouch: HTTP2::Error::ProtocolError
```

this happens due to this code [here](https://github.com/igrigorik/http-2/blob/master/lib/http/2/connection.rb#L455-L457) (after calling `goaway` on a connection, `@state` is marked as `:closed`). 

I was looking into the RFC, and couldn't figure out if this was a strict check besides `An endpoint MUST NOT send frames other than PRIORITY on a closed stream.`, which in my case the sender doesn't know, as the goaway frame has only been buffered, not sent yet. On the other hand, this only happens because I'm draining the socket and reacting on every frame, instead of draining socket **then** doing stuff like closing the connection. Is this something that the library could potentially figure out?

HoneyryderChuck:
Will fix #139 .

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -94,6 +94,7 @@ module HTTP2
       @error = nil
 
       @h2c_upgrade = nil
+      @closed_since = nil
     end
 
     def closed?
@@ -144,6 +145,7 @@ module HTTP2
       send(type: :goaway, last_stream: last_stream,
            error: error, payload: payload)
       @state = :closed
+      @closed_since = Time.now
     end
 
     # Sends a WINDOW_UPDATE frame to the peer.
@@ -446,12 +448,15 @@ module HTTP2
           # the connection, although a new connection can be established
           # for new streams.
           @state = :closed
+          @closed_since = Time.now
           emit(:goaway, frame[:last_stream], frame[:error], frame[:payload])
         when :altsvc, :blocked
           emit(frame[:type], frame)
         else
           connection_error
         end
+      when :closed
+        connection_error if (Time.now - @closed_since) > 15
       else
         connection_error
       end
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -542,6 +542,13 @@ RSpec.describe HTTP2::Connection do
       expect { @conn.new_stream }.to raise_error(ConnectionClosed)
     end
 
+    it 'should not raise error when receiving connection management frames after closing' do
+      @conn.goaway
+      expect(@conn).to be_closed
+
+      expect { @conn << f.generate(PING.dup) }.not_to raise_error(ProtocolError)
+    end
+
     it 'should process connection management frames after GOAWAY' do
       @conn << f.generate(SETTINGS.dup)
       @conn << f.generate(HEADERS.dup)
--- a/spec/connection_spec.rb
+++ b/spec/connection_spec.rb
@@ -542,7 +542,7 @@ RSpec.describe HTTP2::Connection do
       expect { @conn.new_stream }.to raise_error(ConnectionClosed)
     end
 
-    it 'should not raise error when receiving connection management frames after closing' do
+    it 'should not raise error when receiving connection management frames immediately after emitting goaway' do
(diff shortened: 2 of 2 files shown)


lib/http/2/connection.rb
```diff
@@ -446,12 +448,15 @@ def connection_management(frame)
...
           @state = :closed
+          @closed_since = Time.now
           emit(:goaway, frame[:last_stream], frame[:error], frame[:payload])
         when :altsvc, :blocked
           emit(frame[:type], frame)
         else
           connection_error
         end
+      when :closed
+        connection_error if (Time.now - @closed_since) > 15
```
igrigorik: Overall, what you wired up here makes sense. 👍 

One question I have is.. Current logic will cover {settings, window_update, ping, goaway} connection management frames. What about other types of frames, do we need to do anything there? Per https://github.com/igrigorik/http-2/issues/139#issuecomment-414138927...

> After sending a GOAWAY frame, the sender can discard frames for
> streams initiated by the receiver with identifiers higher than the
> identified last stream. However, any frames that alter connection
> state cannot be completely ignored. For instance, HEADERS,
> PUSH_PROMISE, and CONTINUATION frames MUST be minimally processed to
> ensure the state maintained for header compression is consistent (see
> Section 4.3); similarly, DATA frames MUST be counted toward the
> connection flow-control window. Failure to process these frames can
> cause flow control or header compression state to become
> unsynchronized.

Anything else / special we need to think about with respect to the above?

HoneyryderChuck: I'd say it's already covered [here](https://github.com/igrigorik/http-2/blob/3f0cf83c68d11cd04300f186007a28b132b7f8f4/lib/http/2/connection.rb#L253), [here](https://github.com/igrigorik/http-2/blob/3f0cf83c68d11cd04300f186007a28b132b7f8f4/lib/http/2/connection.rb#L253) or [here](https://github.com/igrigorik/http-2/blob/3f0cf83c68d11cd04300f186007a28b132b7f8f4/lib/http/2/connection.rb#L321-L324).

igrigorik: Fair enough. One corner case here is that we'll be ignoring _all_ frames, irrespective of their ID being established before or after GOAWAY is emitted, but that's probably OK. If that surfaces as an issue, we can revisit.

HoneyryderChuck: That is true, and I thought about that. One should throw protocol error for when a stream id is higher than the last stream id from the goaway frame. But I thought that was out of scope. Am I wrong?

igrigorik: I think it's worth leaving a comment in the source, stating that we're _not_ enforcing this behavior. I don't think we need to block this PR on implementing that though.

spec/connection_spec.rb
```diff
@@ -542,6 +542,13 @@
       expect { @conn.new_stream }.to raise_error(ConnectionClosed)
     end
 
+    it 'should not raise error when receiving connection management frames after closing' do
```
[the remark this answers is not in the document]
HoneyryderChuck: just added.

lib/http/2/connection.rb
HoneyryderChuck: after reading better, I think I advised for something which is not RFC-compliant. I think this PR covers everything necessary.

PR #141, opened 2018-10-04
alt-svc frame origin checks from rfc

HoneyryderChuck:
Added the origin checks from the RFC for the `alt-svc` frame.

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -450,7 +450,14 @@ module HTTP2
           @state = :closed
           @closed_since = Time.now
           emit(:goaway, frame[:last_stream], frame[:error], frame[:payload])
-        when :altsvc, :blocked
+        when :altsvc
+          # 4.  The ALTSVC HTTP/2 Frame
+          # An ALTSVC frame on stream 0 with empty (length 0) "Origin"
+          # information is invalid and MUST be ignored.
+          if frame[:origin] && !frame[:origin].empty?
+            emit(frame[:type], frame)
+          end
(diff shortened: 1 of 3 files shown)


lib/http/2/connection.rb
```diff
@@ -450,7 +450,14 @@ def connection_management(frame)
           @state = :closed
           @closed_since = Time.now
           emit(:goaway, frame[:last_stream], frame[:error], frame[:payload])
-        when :altsvc, :blocked
+        when :altsvc
+          # 4.  The ALTSVC HTTP/2 Frame
+          # An ALTSVC frame on stream 0 with empty (length 0) "Origin"
+          # information is invalid and MUST be ignored.
+          if frame[:origin] && !frame[:origin].empty?
```
igrigorik: For stream you have `.nil?`check, any reason why you omitted it here?

HoneyryderChuck: checking that the frame exists and is not empty was the condition, there was no gain in being explicit with `!frame[:origin].nil?`, IMO.

igrigorik: So why do check for nil in stream.rb? Should we remove it there too?

HoneyryderChuck: I see what you mean. I'm out til Monday, I'll sort it out by then.

HoneyryderChuck: done

igrigorik:
Nope, sorry missed your update earlier. Merging. Thanks!

PR #142, opened 2018-11-13
send connection preface on receive

HoneyryderChuck:
Force connection preface when waiting to receive settings frame from server.

I've been trying to get the same frame output as nghttp, which receives the settings frame from server before sending any frame. This hasn't been supported, as connection preface is only sent when calling `#send`. 

This PR fixes the issue.

I'd also like to get rid of that `alias` call (or `receive` altogether, as all examples use `<<`, this is one example where clever aliases can bite you back).


igrigorik:
Overall makes sense, but (a) we need to fix the failing tests, (b) could we add a test to help cover the new behavior?

HoneyryderChuck:
Will do, although some of those tests don't seem to make much sense anymore.

HoneyryderChuck:
@igrigorik done. IMO one should also ditch `#receive` and just use `<<`, but that'd be a backwards-incompatible change, so the ball's on you.

HoneyryderChuck:
👍 @igrigorik do we have enough for a new release?

igrigorik:
Yeah, I think so.. v0.10.1?

HoneyryderChuck:
No breaking changes that I can think of

igrigorik:
Should be live on rubygems :)

HoneyryderChuck:
Sweet! Thx!

Issue #146, opened 2019-02-26
SETTINGS_HEADER_TABLE_SIZE validation?


cjyclaire:
while there is configuration option for `:settings_header_table_size` https://github.com/igrigorik/http-2/blob/70a8a2eed17091e311322f28a7378bba744ad2a6/lib/http/2/connection.rb#L13

there is no validation upon it for coming frames
https://github.com/igrigorik/http-2/blob/70a8a2eed17091e311322f28a7378bba744ad2a6/lib/http/2/connection.rb#L580

May we have validation on coming frames `:settings_header_table_size` value be lower or equal to the limit set by the SETTINGS_HEADER_TABLE_SIZE parameter?

https://http2.github.io/http2-spec/compression.html#encoding.context.update

HoneyryderChuck:
closing this, as it's fixed in 1.0.0 . The patch is [a bit more involved than the linked one](https://github.com/igrigorik/http-2/blob/main/lib/http/2/header/encoding_context.rb#L157-L166).

PR #147, opened 2019-02-28
Add validation to dynamic header table update

cjyclaire:
addressing #146 
https://http2.github.io/http2-spec/compression.html#encoding.context.update

### DIFF
--- a/lib/http/2/compressor.rb
+++ b/lib/http/2/compressor.rb
@@ -146,6 +146,9 @@ module HTTP2
 
         case cmd[:type]
         when :changetablesize
+          if cmd[:value] > @limit
+            fail ProtocolError, 'dynamic table size update exceed limit'
+          end
           self.table_size = cmd[:value]
 
         when :indexed
--- a/lib/http/2/compressor.rb
+++ b/lib/http/2/compressor.rb
@@ -103,7 +103,6 @@ module HTTP2
         @table = []
         @options = default_options.merge(options)
         @limit = @options[:table_size]
-        @dynamic_table_size_update_limit = options[:settings_header_table_size] || 4096
       end
 
       # Duplicates current compression context
@@ -147,7 +146,7 @@ module HTTP2
 
         case cmd[:type]
         when :changetablesize
-          if cmd[:value] > @dynamic_table_size_update_limit
+          if cmd[:value] > @limit
             fail ProtocolError, 'dynamic table size update exceed limit'
           end
           self.table_size = cmd[:value]
@@ -472,10 +471,6 @@ module HTTP2
       # @param options [Hash] decoding options.  Only :table_size is effective.
       def initialize(**options)
         @cc = EncodingContext.new(options)
-      end
-
-      def dynamic_table_size_update_limit=(size)
-        @cc.dynamic_table_size_update_limit = size
       end
 
       # Set dynamic table size in EncodingContext
--- a/lib/http/2/compressor.rb
+++ b/lib/http/2/compressor.rb
@@ -147,7 +147,7 @@ module HTTP2
         case cmd[:type]
         when :changetablesize
           if cmd[:value] > @limit
-            fail ProtocolError, 'dynamic table size update exceed limit'
+            fail CompressionError, 'dynamic table size update exceed limit'
           end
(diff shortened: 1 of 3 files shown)


cjyclaire:
Hmmm might need to rework on this, I need to make sure the checking with the limit set from : SETTINGS_HEADER_TABLE_SIZE

cjyclaire:
might add more docs later, review ready

lib/http/2/compressor.rb
```diff
@@ -103,6 +103,7 @@ def initialize(**options)
         @table = []
         @options = default_options.merge(options)
         @limit = @options[:table_size]
+        @dynamic_table_size_update_limit = options[:settings_header_table_size] || 4096
```
igrigorik: Why do we need this if @limit is already tracking this?

lib/http/2/compressor.rb
```diff
@@ -146,6 +147,9 @@ def process(cmd)
 
         case cmd[:type]
         when :changetablesize
+          if cmd[:value] > @dynamic_table_size_update_limit
```
igrigorik: Same as above, why can't we use @limit?

lib/http/2/compressor.rb
```diff
@@ -470,6 +474,10 @@ def initialize(**options)
         @cc = EncodingContext.new(options)
       end
 
+      def dynamic_table_size_update_limit=(size)
```
igrigorik: Is this necessary if we're using @limit?

lib/http/2/compressor.rb
cjyclaire: my bad, I confused myself originally, then I realized it was set properly for the limit already

cjyclaire:
@igrigorik  Thanks for the review! I simplified the diff and made sure rubycop passed, mind take a second look?  : D (through jruby is failing still but not related to the changes I made ...)

Apologies, I confused my self originally that why I made it complicated first, I read the spec section again and double check the code base, @limit does satisfied the need already.

lib/http/2/compressor.rb
```diff
@@ -146,6 +146,9 @@ def process(cmd)
 
         case cmd[:type]
         when :changetablesize
+          if cmd[:value] > @limit
+            fail ProtocolError, 'dynamic table size update exceed limit'
```
igrigorik: > A value that exceeds this limit MUST be treated as a decoding error.

WDYT of raising [CompressionError](https://github.com/igrigorik/http-2/blob/master/lib/http/2/error.rb#L16-L19)?

cjyclaire: Sounds good, updated in the latest commit : )

ioquatix:
I ran into this implementation while working through h2spec and my fork of this code in `protocol-hpack`, and found that this implementation seems wrong to me.

The `@limit` is related to the table size, but it's adjusted to be smaller if table size is changed. This doesn't seem right, in my implementation I'm changing `@limit` => `@table_size` and adding `@table_size_limit` which is set independently.

Just FYI.

ioquatix:
I've been playing around with this a bit more, in order to try and be compliant with RFC, but I don't completely understand how we should handle settings relating to table size.

>    SETTINGS_HEADER_TABLE_SIZE (0x1):  Allows the sender to inform the
>       remote endpoint of the maximum size of the header compression
>       table used to decode header blocks, in octets.  The encoder can
>       select any size equal to or less than this value by using
>       signaling specific to the header compression format inside a
>       header block (see [COMPRESSION]).  The initial value is 4,096
>       octets.

So, if we receive this on the server, when we encode headers (e.g. for response headers), we should limit the dynamic table to the specified size?

ioquatix:
Okay, so I found a potential problem with the implementation.

Essentially, if you change the table size, you need to emit the right `:change_table_size` command. But right now the implementation doesn't appear to do it. So, I implemented it.

https://github.com/socketry/protocol-hpack/commit/51879fbea2ea205aba3fed59072a94acaca04d49

It was pretty straight forward to add once I understood how it was supposed to be used. My implementation is pretty naive, it just changes the table size to the maximum the client allows. I assume this is resource intensive.

PR #152, opened 2020-05-25
Fix cleaning up recently closed streams

crazyshot:
Ever since 16dec9f2c44d3a525a49fe92bc15c7b79e9de097 in 2018
@streams_recently_closed cleanup was broken.

`to_delete` contains hash of { stream_id => time } so in
`to_delete.each do |stream_id|` `stream_id` is actually array of `[
stream_id, time ]`.

This leads to the same memory leak as there was before the first cleanup
but also adds heavy CPU usage when thousands of streams are being
checked on every stream close.

Since 15s delayed cleanup was intended but actually infinite delay was
added so 15s was never tested in production code and thus might be not
enough.

This change also improves performance of searching streams to delete by
relying on the fact that Ruby Hash is ordered and so once fresh stream
was met we can stop searching.

Also minor performance and memory improvement by storing only timestamp
instead of Time object.

Extracted this part of the code to a separate function for test
convenience.

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -33,6 +33,9 @@ module HTTP2
   # Default connection "fast-fail" preamble string as defined by the spec.
   CONNECTION_PREFACE_MAGIC = "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n".freeze
 
+  # Time to hold recently closed streams until purge (seconds)
+  RECENTLY_CLOSED_STREAMS_TTL = 15
+
   # Connection encapsulates all of the connection, stream, flow-control,
   # error management, and other processing logic required for a well-behaved
   # HTTP 2.0 endpoint.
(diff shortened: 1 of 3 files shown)


crazyshot:
Actually I think that the cleanup shouldn't be related to the `:close` event and should be triggered by an external timer every second, this way even if there won't be any stream close it will still do the cleanup. But I wasn't sure it's safe to do it with multithreading and not sure it can be done via already existing Emitter events. Maybe `on(:frame)` with time throttle could do the trick?

lib/http/2/connection.rb
```diff
@@ -698,6 +697,21 @@ def activate_stream(id: nil, **args)
       @streams[id] = stream
     end
 
+    # Purge recently streams closed within defined RTT time window.
+    def cleanup_recently_closed
+      now_ts = Time.now.to_i
+      to_delete = []
+      @streams_recently_closed.each do |stream_id, ts|
```
igrigorik, line 704: > This change also improves performance of searching streams to delete by relying on the fact that Ruby Hash is ordered and so once fresh stream was met we can stop searching.

Can you add this as a comment? Good thing to document for future reference.

crazyshot:
Hi @igrigorik, will you merge it?

igrigorik:
Thanks again for the patch!

crazyshot:
@igrigorik Hi. Do you have a plan when to release a new version?

igrigorik:
Sorry about the delay, 0.10.2 should be live now!

PR #154, opened 2020-06-17
Test with Ruby 2.5-2.7, cache bundler in CI

aried3r:
`sudo: false` is deprecated.


aried3r:
This also reveals some deprecation warnings of Ruby 2.7, but I'd rather have fixes for those as a separate PR and for now make sure newest Rubies are being tested, despite the warnings.

igrigorik:
Thank you!

Issue #151, opened 2019-10-10
ignoring frame size negotiation


HoneyryderChuck:
Coming from: https://gitlab.com/honeyryderchuck/httpx/issues/59 .

If you look at the shown frame negotiation, you can see that the server negotiates an increase of max frame size. However, this is never taken into account, as data frame parsing fails with `common_header': Frame size is too large: 95846 (HTTP2::Error::CompressionError)`.

The main offender seems to be this condition, which doesn't do a thing.

igrigorik:
Resolved via #157.

PR #157, opened 2020-11-11
update max frame size on receiving SETTINGS frame from peer

sankarcheppali:
Currently the framer max_frame_size is not being updated as reported in [151](https://github.com/igrigorik/http-2/issues/151).  This PR tries fix that

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -539,7 +539,7 @@ module HTTP2
         # Process pending settings we have sent.
         [@pending_settings.shift, :local]
       else
-        connection_error(check) if validate_settings(@remote_role, frame[:payload])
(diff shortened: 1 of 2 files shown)


igrigorik (commented):
Thank you for the contribution!  Overall, looks great. 

Q: should there be a min/max boundary check?

sankarcheppali:
> Thank you for the contribution! Overall, looks great.
> 
> Q: should there be a min/max boundary check?

Cool. We are already doing a check with validate_settings before accepting the remote settings. We could add one more check in framer if you feel that is required

igrigorik:
Ah, good point! No, should be good. Ready to merge, or do you have anything else pending?

sankarcheppali:
We are good to merge 👍

PR #161, opened 2022-09-14
Support running with --enable=frozen-string-literal

danielmorrison:
In apps that run with `--enable=frozen-string-literal` we can get a `FrozenError`.

Using `String.new` instead of `''` fixes that without breaking it for anyone else.

When running the specs frozen strings (`RUBYOPT="--enable=frozen-string-literal" rake`), there were two files that have errors. I added the magic comment: `# frozen-string-literal: false` to those to work both ways.


igrigorik (approved):
👍 thanks for the contribution!

PR #164, opened 2024-05-31
Enable build with GH actions

mullermp:


mullermp:
Hm. This did not do the trick. Let me try to open a PR on the fork first.

mullermp:
The CI runs correctly on https://github.com/mullermp/http-2/pull/1 so I'm not sure why it's not starting here - maybe a permissions thing?

The failures are from old Ruby. I would like to submit a different PR that bumps the minor version and sets minimum ruby to at least 2.3 (or even 2.5). Let me know if you're ok with that.

flavorjones:
Just a note that, in Github Actions, a workflow file must exist on the default branch before it will run in pull requests.

mullermp:
Makes sense for security reasons! Thank you.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - uses: actions/checkout@v4
+
+      - name: Install gems
+        run: bundle install
```
alextwoods, line 26: Could use bundler-cache with setup-ruby, eg: https://github.com/smithy-lang/smithy-ruby/pull/198

alextwoods (approved):

igrigorik (approved):

PR #165, opened 2024-05-31
Organize directory based on module name

mullermp:
The module name is HTTP2, so flatten the directories and paths as such.

I've also moved the private task into a tasks folder (should not be part of lib I think).

Let me know if I should regenerate the huffman table? I'm not entirely sure what this is for - but it does produce different results.


lib/http/2.rb
alextwoods, line 1: I think this would be breaking for everyone using the gem with `require 'http/2''

mullermp:
With the module name as `HTTP2`, but the gem as `http-2`, and following gem naming standards, there are two approaches:

1. Rename module to HTTP::2 and keep current directory, or
2. Rename the folders as I've done, but doesn't comply well with gem name.

Both options can be fixed with a major version bump or by aliasing. I actually maybe think I should go with option 1 because it aligns with the gem name (should not change at all).

mullermp:
We can either keep the alias or major version bump with a minimum Ruby version.

lib/http2.rb
```diff
@@ -0,0 +1,13 @@
...
+require 'http2/error'
+require 'http2/emitter'
+require 'http2/buffer'
+require 'http2/flow_buffer'
+require 'http2/huffman'
+require 'http2/huffman_statemachine'
+require 'http2/compressor'
+require 'http2/framer'
+require 'http2/connection'
+require 'http2/client'
```
alextwoods, line 11: This change will also break anyone that was using `require 'http/2/client`.  My guess is there isn't much usage like that since the README examples all have the top level file being required and it requires all other files.

mullermp: I don't see this in the SDK so I think we are fine. But you're right that someone could be doing this. Documented usage is different as you point out. I think it's fine personally.

igrigorik (commented):
Hmm, on the fence this one. 

I agree that current code mixes module name convention vs directory structure, but is there a big benefit to imposing this as a breaking change? My intuition is that it's purely aesthetic and not worth the imposed work for everyone to adjust & upgrade?

lib/http2.rb
mullermp: Seeing as this is not expected behavior, this one doesn't worry me, but I'll leave it to the author.

PR #167, opened 2024-06-04
Refactor generate_huffman_table task

mullermp:
Pulls this task out of lib and regenerate the table.


mullermp:
@igrigorik Can you tell me if this makes sense to do?

mullermp:
I did a local install of this gem and tested it with CloudWatchLogs start_live_tail and Transcribe Streaming (both http2) and it appears to work.

mullermp:
Regenerating the table seems dangerous. Someone with more experience in this area should probably take a look.

alextwoods (approved):

PR #169, opened 2024-06-10
Add base64 as a dependency for Ruby 3.4

mullermp:
Adds base64 as a dependency as required by Ruby 3.4

### DIFF
--- a/http-2.gemspec
+++ b/http-2.gemspec
@@ -11,5 +11,7 @@ Gem::Specification.new do |spec|
   spec.license       = 'MIT'
   spec.files         = Dir["LICENSE", "README.md", "lib/**/*.rb"]
 
+  spec.add_runtime_dependency 'base64'
+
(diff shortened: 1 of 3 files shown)


http-2.gemspec
```diff
@@ -11,5 +11,7 @@ Gem::Specification.new do |spec|
   spec.license       = 'MIT'
   spec.files         = Dir["LICENSE", "README.md", "lib/**/*.rb"]
 
+  spec.add_runtime_dependency 'base64'
```
alextwoods, line 14: Should we do this in a different minor version from the other fixes? (in particular I'm thinking, if this causes issues for some users, being able to use the smaller version of the gem would be useful).

mullermp: I think it's fine to not do that. The current experience is already what's expected, and a new version can be a carrot to upgrade (using new ruby, smaller gem, etc..)

alextwoods (approved):

PR #172, opened 2024-06-12
Port in http-2-next

mullermp:
This has `http-2-next` as is.

WIP to change RBS, documentation, namespaces, etc, to use HTTP2 instead of HTTP2Next.


mullermp:
Closing this, this is just for reviewing diff - it should not be merged into master. Master will need to be renamed to version-0 or tagged with the latest version.

README.md
```diff
@@ -1,16 +1,20 @@
-# HTTP-2
+# HTTP-2-Next
```
mullermp, line 1: @HoneyryderChuck You'd be best to update this file, with links to the correct/new documentation and usage.

lib/http/2/base64.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+if RUBY_VERSION < "3.3.0"
+  require "base64"
+elsif !defined?(Base64)
+  module HTTP2
+    # require "base64" will not be a default gem after ruby 3.4.0
```
mullermp, line 7: I think adding base64 as a dependency to the gem will make this file obsolete.

igrigorik (commented):
A substantial part of the diff here is due to change from literal to interpolated strings.. i.e. ""s instead of ''s. Why, though? Can we unwind this, there is no benefit and overhead to interpolated strings that we do not need.

HoneyryderChuck:
> why though?

It was not the only rubocop rule change prior to the fork. Why i did it though: supporting single and double quotes makes using strings needlessly ambiguous and unreadable. It's a part of the default rubocop config that the standard gem didn't inherit, for example.

I think this PR here is only for tracking purposes, as the goal is to use this branch as the default one. From that perspective, it's better to keep conservative to the imported changes (an attempt to reformat things last week caused a regression in tests) rather than aligning with the master branch. I'm assuming, of course, there's no other benefit in retrofitting single quotes (do correct me if there is one).

mullermp:
My understanding is that it's a best practice and also slightly faster to use literals. It would allow for the diff to be readable, yes, and I highly recommend single quotes here. There should not be any regressions and I'm pretty sure I had all the tests passing with single quotes. It is likely an issue in the test with frozen string literal.

mullermp:
I pushed the change with single quoted strings. The tests all pass. Can you tell me what regression to you saw?

While we don't intend to merge the PR, I definitely think we should have string literals. We can keep many of the stylings of your preference but the rubocop defaults are fairly sane.

HoneyryderChuck:
> My understanding is that it's a best practice and also slightly faster to use literals.

Double-quote strings are also literals. [there is no performance difference between them, they're internally the same](https://stackoverflow.com/questions/1836467/is-there-a-performance-gain-in-using-single-quotes-vs-double-quotes-in-ruby).

It's not a best practice, it's just rubocop default config. The inconsistency of the style guide created by the rubocop creator is well documented, and one of the outcomes is [a very popular gem which is essentially rubocop with different defaults](https://github.com/standardrb/standard). 

> There should not be any regressions and I'm pretty sure I had all the tests passing with single quotes.

I was talking about your first attempt porting http-2-next here last week, which I found a regression for when working on porting gitlab CI to github actions (it's in our slack conversation).

@mullermp  Bear in mind that we agreed [here](https://github.com/igrigorik/http-2/pull/144#issuecomment-2160365599) on doing a hard-merge, and I mentioned last week to you on slack as well that I'd be open to backtrack on some of the styling changes I committed after the fork (although I'd rather do it after there's a release), but I'd rather keep this one about using double-quoted strings, for the stated reasons, to which there was no objection. I appreciate @igrigorik hasn't joined the chat yet, but I was under the impression you were keeping him in the loop on our conversation.

igrigorik (approved):
LGTM 👍 

Lots of good improvements here, thank you @HoneyryderChuck!

mullermp:
Thanks everyone. I'll update HEAD the be main and close this PR. We should move `master` to be version-0 as an archive.

PR #177, opened 2024-11-04
Improve ractor safety

HoneyryderChuck:
Also, replacing usage of refinements.


lib/http/2/header/huffman_statemachine.rb
```diff
@@ -268,7 +268,7 @@ class Huffman
         [[28, 29], [28, 5], [29, 29], [29, 5], [30, 29], [30, 5], [31, 29], [31, 5], [127, 29], [127, 5], [220, 29], [220, 5], [249, 29], [249, 5], [nil, 254], [nil, 255]],
         [[10, 17], [10, 18], [10, 19], [10, 20], [10, 21], [10, 22], [10, 23], [10, 7], [13, 17], [13, 18], [13, 19], [13, 20], [13, 21], [13, 22], [13, 23], [13, 7]],
         [[22, 17], [22, 18], [22, 19], [22, 20], [22, 21], [22, 22], [22, 23], [22, 7], [256, 17], [256, 18], [256, 19], [256, 20], [256, 21], [256, 22], [256, 23], [256, 7]],
-      ].each { |arr| arr.each { |subarr| subarr.each(&:freeze) }.freeze }.freeze
+      ].each { |arr| arr.each { |subarr| subarr.freeze }.freeze }.freeze
```
mullermp, line 271: Nice. This looks like it was a bug right? Only the arrays need to be frozen not values?

lib/http/2/header/encoding_context.rb
```diff
@@ -77,10 +77,11 @@ class EncodingContext
...
 
-      STATIC_TABLE_BY_FIELD = STATIC_TABLE
-                              .each_with_object({})
-                              .with_index { |((field, value), hs), idx| (hs[field] ||= []) << [idx, value] }
-                              .each { |pair| pair.each(&:freeze).freeze }.freeze
+      STATIC_TABLE_BY_FIELD =
+        STATIC_TABLE
+        .each_with_object({})
+        .with_index { |((field, value), hs), idx| (hs[field.freeze] ||= []) << [idx, value.freeze].freeze }
+        .each { |pair| pair.each(&:freeze).freeze }.freeze
```
mullermp, line 84: Same thing here right? Do pair values need to be frozen or only the arrays? There are probably instances of this in other places.

lib/http/2/header/encoding_context.rb
```diff
@@ -77,10 +77,11 @@ class EncodingContext
...
       ].each { |pair| pair.each(&:freeze).freeze }.freeze
 
-      STATIC_TABLE_BY_FIELD = STATIC_TABLE
-                              .each_with_object({})
-                              .with_index { |((field, value), hs), idx| (hs[field] ||= []) << [idx, value] }
-                              .each { |pair| pair.each(&:freeze).freeze }.freeze
+      STATIC_TABLE_BY_FIELD =
+        STATIC_TABLE
+        .each_with_object({})
+        .with_index { |((field, value), hs), idx| (hs[field.freeze] ||= []) << [idx, value.freeze].freeze }
```
mullermp, line 83: Ditto on freeze question

mullermp (approved):
Nice clean up. Just added some considerations, it's otherwise fine.

lib/http/2/header/huffman_statemachine.rb
HoneyryderChuck: exactly. not a "bug" as in, it didn't lead to breaking code, but it made this constant unusable across ractors.

lib/http/2/header/encoding_context.rb
HoneyryderChuck: indeed, arrays must be frozen to be shareable (values already were).

lib/http/2/header/encoding_context.rb
HoneyryderChuck: same

lib/http/2/header/encoding_context.rb
mullermp: I'm essentially asking if you need `pair.each(&:freeze)` here? I would assume you don't want values frozen.

lib/http/2/header/encoding_context.rb
mullermp: Do you need `value.freeze`? I thought just arrays needed to be.

lib/http/2/header/encoding_context.rb
HoneyryderChuck: you're right, I've simplified it now, and not try to freeze values that I know are already frozen.

PR #179, opened 2025-03-26
adding release workflow for github actions to enable trusted publishing

HoneyryderChuck:
this should allow users with publish permissions to "review and merge" while creating a new version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
       - master
 
 env:
-  ruby_version: 3.3
+  ruby_version: 3.4
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


mullermp (commented):
I'm going to assume you followed the onboarding steps.

mullermp (approved):

HoneyryderChuck:
Not yet, buy I will 👍 

One thing that is not clear to me yet is how this will be activated. The plugin readme mentions the rake release tasks, but does not specify which (and I assume it does not mean that I should run the `rake release` task this flow is supposed to replace. So, should I then push the release tag with the branch open? Would be great to have feedback from someone who experienced it.

PR #181, opened 2025-04-03
v1.1.0

HoneyryderChuck:
a few last minute signature improvements, and the version + changelog changes.

Testing Trusted Publishing as well through it.

### DIFF
--- a/lib/http/2/version.rb
+++ b/lib/http/2/version.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module HTTP2
-  VERSION = "1.0.2"
+  VERSION = "1.1.0"
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 1.1.0
+
+Several changes which improved performance for the common cases. A few highlights:
+
+* Opting into ruby 3.4 features when possible (such as `String#append_as_bytes` instead of `String#<<`)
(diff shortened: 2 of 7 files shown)


lib/http/2/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module HTTP2
-  VERSION = "1.0.2"
+  VERSION = "1.1.0"
```
mullermp, line 4: I think version gets bumped automatically with the publisher, doesn't it? I thought I read something about version.rb being assumed.

mullermp (approved):

lib/http/2/version.rb
HoneyryderChuck: it doesn't (and I'm counting on it, as changes in this file trigger the release workflow).

HoneyryderChuck: https://github.com/rubygems/release-gem/blob/v1/action.yml

mullermp: https://bundler.io/guides/creating_gem.html#releasing-the-gem

This is what I was thinking of. I guess it's gem bump that could do this. So that can be in the CI if you wanted.

HoneyryderChuck: Not sure if that's the most feasible. You need a way to tell CI that you want to bump the version, and whether it's a patch/minor/major bump. Bumping locally and triggering the release flow on changes to version.rb seems more straightforward.

PR #183, opened 2025-04-14
Buffer encoding issues

HoneyryderChuck:
Contains fixes for some string encoding issues reported [here](https://github.com/HoneyryderChuck/httpx/issues/83) and [here](https://gitlab.com/os85/httpx/-/issues/345).

Releases a new patch version (1.1.1)

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -31,7 +31,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 29
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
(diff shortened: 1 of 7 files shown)


mullermp (approved):
Seems fine - I can do a sanity test with the SDK later this week.

HoneyryderChuck:
thx. I'll be issuing a release today though, as this was disruptive to some folks who upgraded the lib, and I'd expect more to come as time passes by. I already ran a few sanity checks on my side as well, and the original issue reporters have already deployed the patch and confirmed it fixes the issue.

PR #191, opened 2026-07-04
Fix bytesplice

HoneyryderChuck:
fix String#bytesplice usage, which in order to be able to cut UTF-8 characters, requires the string to be encoded as binary, otherwise it'll raise a range error.

most of the strings passed to #read_str come from downstream usage of itself, so there are more chances of reducing string creation by returning binary strings.


HoneyryderChuck:
@igrigorik can you go to "Settings" -> "Actions" -> "General" and enable "
Allow all actions and reusable workflows" under "Actions permissions"? Without this, the coverage step won't work. I'd do it myself, but I don't have any settings available for edition (I'd probably need to be an admin of the repo).

PR #192, opened 2026-07-09
Deliver a promised stream's response HEADERS received after GOAWAY

const86:
On receiving GOAWAY the connection moves to :closed, after which #receive dropped every inbound HEADERS frame - including the response HEADERS of a stream the peer promised to finish (id <= last_stream_id). That stream's DATA frames were still delivered (the DATA branch has no :closed guard), so the consumer saw a malformed, headerless response and the request was lost.

RFC 9113 Section 6.8 says a receiver of GOAWAY may still complete streams at or below last_stream_id. Only discard HEADERS that would open a NEW stream (id not already in @streams); keep delivering an in-flight stream's response.

This bites whenever a server emits GOAWAY just ahead of the response on the same connection - e.g. golang.org/x/net/http2 graceful shutdown - which loses responses the server actually processed and put on the wire.

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -300,7 +300,7 @@ module HTTP2
             # still complete (RFC 9113 Section 6.8). Only discard HEADERS that
             # would open a NEW stream; a known stream's response must be
             # delivered (its DATA frames already are - see the branch below).
-            return if @state == :closed && !@streams.key?(stream_id)
+            return if @state == :closed && stream_id > @last_stream_id
 
             stream = @streams[stream_id]
             if stream.nil?
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -295,7 +295,12 @@ module HTTP2
             # PUSH_PROMISE and CONTINUATION frames MUST be minimally
             # processed to ensure a consistent compression state
             decode_headers(frame)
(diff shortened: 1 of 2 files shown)


lib/http/2/connection.rb
```diff
@@ -295,7 +295,12 @@ def receive(data)
             # PUSH_PROMISE and CONTINUATION frames MUST be minimally
             # processed to ensure a consistent compression state
             decode_headers(frame)
-            return if @state == :closed
+            # A GOAWAY moves the connection to :closed, but streams the peer
+            # promised to finish (id <= last_stream_id, already in @streams) can
+            # still complete (RFC 9113 Section 6.8). Only discard HEADERS that
+            # would open a NEW stream; a known stream's response must be
+            # delivered (its DATA frames already are - see the branch below).
+            return if @state == :closed && !@streams.key?(stream_id)
```
HoneyryderChuck:
```suggestion
            return if @state == :closed && stream_id > @last_stream_id
```

HoneyryderChuck (commented):
@const86 thx for the patch! LMK what you think about my suggestion, other than that LGTM

lib/http/2/connection.rb
const86: Works equally well for my cases. Thanks!

PR #193, opened 2026-07-14
Drain in-flight streams after a graceful GOAWAY

const86:
On a graceful GOAWAY the connection moved straight to `:closed`, where connection-level WINDOW_UPDATE and SETTINGS are ignored — and escalate to a connection error 15s later (#140) — while the peer keeps completing streams at or below last_stream_id (RFC 9113 Section 6.8). In-flight uploads stall on the frozen connection window, and drains longer than 15s kill every promised stream.

A graceful GOAWAY received while streams are tracked now enters `:closing`: connection-level frames are processed as usual, tracked streams run to completion, and the connection moves to `:closed` when the last one ends (or immediately on an error GOAWAY). HEADERS that would open a new stream and PUSH_PROMISE are discarded (after HPACK decoding, keeping compression state consistent), and `new_stream` keeps raising ConnectionClosed.

`closed?` now answers "does this connection accept new work?" — true from GOAWAY receipt on, exactly when it turned true before this change; `closing?` distinguishes a still-draining connection. Streams the GOAWAY refused (or that never started) hold nothing open: callers keep tracking refusals via the `:goaway` event and retry on a new connection.

After a GOAWAY, a PRIORITY frame for an already-used stream id no longer resurrects a phantom stream (previously it re-activated the id and emitted `:stream` with no HEADERS behind it — while draining, such a phantom would also hold the connection in `:closing` forever). Live-connection PRIORITY handling is unchanged.

Draining only affects frames the application feeds via `#receive` — nothing is buffered behind its back (the #184 concern). Subsumes the #192 HEADERS loss while draining; #192 still applies to an already-closed connection.

Validated end-to-end against golang.org/x/net/http2 graceful shutdown (h2c): session streams keep serving mid-drain, a 3 MB upload crosses the connection window while :closing, and post-GOAWAY stream opens fail fast for retry on a fresh connection.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/http/2/connection.rb
+++ b/lib/http/2/connection.rb
@@ -56,7 +56,8 @@ module HTTP2
     include Error
     include BufferUtils
 
-    # Connection state (:new, :closed).
+    # Connection state (:waiting_magic, :waiting_connection_preface,
+    # :connected, :closing, :closed).
     attr_reader :state
 
     # Size of current connection flow control window (by default, set to
@@ -113,8 +114,15 @@ module HTTP2
       @send_buffer = FrameBuffer.new
     end
 
+    # No new streams may be opened (a GOAWAY was sent or received).
     def closed?
-      @state == :closed
+      @state == :closed || @state == :closing
+    end
+
+    # Graceful GOAWAY received: tracked streams may still complete,
+    # but no new ones can be opened.
+    def closing?
+      @state == :closing
     end
 
     # Allocates new stream for current connection.
@@ -123,7 +131,7 @@ module HTTP2
     # @param window [Integer]
     # @param parent [Stream]
     def new_stream(**args)
-      raise ConnectionClosed if @state == :closed
+      raise ConnectionClosed if closed?
       raise StreamLimitExceeded if @active_stream_count >= @remote_settings[:settings_max_concurrent_streams]
 
       connection_error(:protocol_error, msg: "id is smaller than previous") if @stream_id < @last_stream_id
@@ -158,8 +166,7 @@ module HTTP2
     def goaway(error = :no_error, payload = nil)
       send(type: :goaway, stream: 0, last_stream: @last_stream_id,
            error: error, payload: payload)
-      @state = :closed
-      @closed_since = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+      close!
     end
 
     # Sends a WINDOW_UPDATE frame to the peer.
@@ -295,12 +302,11 @@ module HTTP2
             # PUSH_PROMISE and CONTINUATION frames MUST be minimally
             # processed to ensure a consistent compression state
             decode_headers(frame)
-            # A GOAWAY moves the connection to :closed, but streams the peer
-            # promised to finish (id <= last_stream_id, already in @streams) can
-            # still complete (RFC 9113 Section 6.8). Only discard HEADERS that
-            # would open a NEW stream; a known stream's response must be
-            # delivered (its DATA frames already are - see the branch below).
-            return if @state == :closed && stream_id > @last_stream_id
+            # After a GOAWAY, HEADERS may only complete a tracked stream at
+            # or below the highest id seen (Section 6.8); anything else is
+            # discarded - decoded above, so the compression state is intact.
+            # Skip only this frame: later ones may serve completing streams.
+            next if closed? && (stream_id > @last_stream_id || !@streams.key?(stream_id))
 
             stream = @streams[stream_id]
             if stream.nil?
@@ -328,7 +334,7 @@ module HTTP2
             end
 
             decode_headers(frame)
-            return if @state == :closed
+            next if closed?
 
             # PUSH_PROMISE frames MUST be associated with an existing, peer-
             # initiated stream... A receiver MUST treat the receipt of a
@@ -385,6 +391,10 @@ module HTTP2
               # group of dependent streams by altering the priority of an
               # unused or closed parent stream.
               when :priority
+                # The stream already existed and closed: reprioritizing a
+                # closed parent is a no-op here - do not resurrect it.
+                next if stream_id <= @last_stream_id
+
                 stream = activate_stream(
                   id: stream_id,
                   weight: frame[:weight] || DEFAULT_WEIGHT,
@@ -500,7 +510,7 @@ module HTTP2
         @state = :connected
         connection_settings(frame)
 
-      when :connected
+      when :connected, :closing
         case frame_type
         when :settings
           # @type var frame: settings_frame
@@ -515,8 +525,13 @@ module HTTP2
           # Receivers of a GOAWAY frame MUST NOT open additional streams on
           # the connection, although a new connection can be established
           # for new streams.
-          @state = :closed
-          @closed_since = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          if frame[:error] == :no_error && !@streams.empty?
+            # A graceful GOAWAY lets tracked streams complete before the
+            # connection closes (Section 6.8).
+            @state = :closing
+          else
+            close!
+          end
           emit(:goaway, frame[:last_stream], frame[:error], frame[:payload])
         when :altsvc
           # @type var frame: altsvc_frame
@@ -786,6 +801,8 @@ module HTTP2
       stream.once(:close) do
         @streams.delete(id)
 
+        close! if @state == :closing && @streams.empty?
+
         # Store a reference to the closed stream, such that we can respond
         # to any in-flight frames while close is registered on both sides.
         # References to such streams will be purged whenever another stream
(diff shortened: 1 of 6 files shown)


lib/http/2/connection.rb
```diff
@@ -117,13 +117,19 @@ def closed?
       @state == :closed
     end
 
+    # The connection received a graceful GOAWAY: tracked streams may still
+    # complete, but no new streams can be opened.
+    def draining?
```
HoneyryderChuck: can we rename it to `closing?` ?

lib/http/2/connection.rb
```diff
@@ -811,6 +835,15 @@ def activate_stream(id:, **args)
       @streams[id] = stream
     end
 
+    # Streams the peer may still complete after a graceful GOAWAY
+    # (Section 6.8): already begun (neither idle nor reserved) and at or
+    # below the advertised last stream id.
+    def streams_to_drain?(last_stream)
```
HoneyryderChuck: this check is quite expensive. I feel it's a bit unnecessary though, since connections do not allow new streams after a last stream is set, so all you need to do IMO is to check whether `@streams.empty?`. WDYT?

const86: Agreed — but bare `@streams.empty?` would hang on streams that never complete (beyond last_stream_id, or idle/reserved). The GOAWAY handler now drops those up front, so `@streams.empty?` is the drained condition and `streams_to_drain?` is gone.

lib/http/2/connection.rb
```diff
@@ -328,7 +336,7 @@ def receive(data)
             end
 
             decode_headers(frame)
-            return if @state == :closed
+            next if @state == :closed || @state == :closing
```
HoneyryderChuck: judging my the amount of times I'm seeing `@state == :closed || @state == :closing`, I wonder whether it'd be best to add it to the `closed?` method and reuse it everywhere.

lib/http/2/connection.rb
```diff
@@ -385,6 +393,10 @@ def receive(data)
               # group of dependent streams by altering the priority of an
               # unused or closed parent stream.
               when :priority
+                # An idle stream activated here would never close, keeping a
```
HoneyryderChuck: I'm not sure this is correct. according to the comment, PRIORITY frames can be received in the idle or closed stream state. besides, `activate_stream` will fail if the stream hasn't been created yet, and reprocessing of priority, while it hasn't been implemented yet, may be so in the future. Should be best to keep it as is, don't you think?

There is a bug here though: closed streams won't be available in `@streams`, so this block will raise an error for closed streams (or worse, create them again?). Therefore, this block should be skipped **if** stream_id is lower than @last_stream_id and there's no stream in `@streams` (meaning, it has already been closed). Feel free to give it a go at implementing this, otherwise I can do it post-merge.

lib/http/2/stream.rb
```diff
@@ -300,6 +300,17 @@ def refuse
       send(type: :rst_stream, error: :refused_stream)
     end
 
+    # Closes the stream locally, without RST_STREAM, once a GOAWAY says the
+    # peer will not process it: refused (id beyond last_stream_id) or idle
+    # (RST_STREAM on idle is forbidden, Section 6.4).
+    def close_locally(error = :refused_stream)
```
HoneyryderChuck: besides thinking that this is unnecessary, it may also be confusing, as it shouldn't be called when in certain states (that condition is held by the branch in the connection when calling this function).

lib/http/2/connection.rb
```diff
@@ -515,9 +527,25 @@ def connection_management(frame)
           # Receivers of a GOAWAY frame MUST NOT open additional streams on
           # the connection, although a new connection can be established
           # for new streams.
-          @state = :closed
-          @closed_since = Process.clock_gettime(Process::CLOCK_MONOTONIC)
           emit(:goaway, frame[:last_stream], frame[:error], frame[:payload])
+          if frame[:error] == :no_error
```
HoneyryderChuck: IMO this whole block is pointless, besides it being expensive to process. It'd also fail for idle streams receiving a priority frame (although my suggestion above would fix that condition). The reason I think it's pointless is because, if a stream object is found, it'll only be taken into account if receiving DATA frames. Other than, they'd be ignored already. I think that this part of the spec matters only **if** the connection object were reusable, which it isn't (retries will happen in a new connection).

lib/http/2/connection.rb
```diff
@@ -515,9 +527,25 @@ def connection_management(frame)
...
+            doomed = @streams.each_value.select do |stream|
+              stream.id > frame[:last_stream] ||
+                STREAM_UNOPENED_STATES.include?(stream.state)
+            end
+            doomed.each(&:close_locally)
+          end
+          if frame[:error] == :no_error && !@streams.empty?
+            @state = :closing
+          else
+            @state = :closed
```
HoneyryderChuck: these instructions are now repeated so many times that it's probably worth moving them to a private method (like `#close`? `#close!`? up to you)

lib/http/2/connection.rb
const86: `closed?` includes `:closing`, so it still turns true at GOAWAY receipt — same timing as before this PR. `closing?` tells a draining connection apart.

lib/http/2/connection.rb
const86: Per §5.1.1 an untracked id ≤ `@last_stream_id` is already "closed" — nothing to reprioritize. Fresh ids still activate, even while closing; such an idle ghost blocks the terminal `:closed`, harmlessly (`closed?` is already true).

lib/http/2/connection.rb
const86: Dropped. Refused/idle leftovers now keep the connection in `:closing` instead of reaching `:closed` — `closed?` is true either way, and a clean drain still ends in `:closed`.

lib/http/2/connection.rb
```diff
@@ -385,6 +391,10 @@ def receive(data)
               # group of dependent streams by altering the priority of an
               # unused or closed parent stream.
               when :priority
+                # The stream already existed and closed: reprioritizing a
+                # closed parent is a no-op here - do not resurrect it.
+                next if stream_id <= @last_stream_id
```
HoneyryderChuck:
```suggestion
                next if closed? && stream_id <= @last_stream_id
```

streams can be initiated by a PRIORITY frame.

lib/http/2/connection.rb
```diff
@@ -786,6 +801,8 @@ def activate_stream(id:, **args)
       stream.once(:close) do
         @streams.delete(id)
 
+        close! if @state == :closing && @streams.empty?
```
HoneyryderChuck, line 808: 💅 just add a comment about why this is necessary.

spec/shared_examples/connection.rb
```diff
@@ -165,6 +165,32 @@
...
 
+    it "should drain in-flight streams after receiving a graceful GOAWAY" do
+      stream = connected_conn.new_stream
+      stream.send headers_frame
+
+      connected_conn << f.generate(goaway_frame)
+      expect(connected_conn).to be_closing
+
+      # Connection-level frames keep being processed while closing.
+      connected_conn << f.generate(window_update_frame.merge(stream: 0, increment: 1000))
```
HoneyryderChuck, line 176: shouldn't this also be testing that frames for `stream` do not raise?

spec/client_spec.rb
```diff
@@ -339,6 +339,207 @@
...
+      finished.close # locally reset: no longer tracked
+
+      survivor = client.new_stream
+      survivor.send headers_frame.merge(stream: survivor.id)
+
+      client << f.generate(goaway_frame.merge(last_stream: survivor.id))
+
+      # Response HEADERS racing the local reset must not kill the
+      # closing connection.
+      client << f.generate(headers_frame.merge(payload: Compressor.new.encode(RESPONSE_HEADERS)))
```
HoneyryderChuck: I see you're doing it there. would it be possible to test this all in the same `it block`?

lib/http/2/connection.rb
const86: While closing, a resurrected phantom would emit `:stream` after the GOAWAY and — never completing — wedge the drain in `:closing`; the reworked test pins both.

spec/shared_examples/connection.rb
const86: DATA rather than HEADERS or a stream WINDOW_UPDATE: the example also runs as a server, where HEADERS on the even stream id is a connection error, and the helper SETTINGS pin windows at 2³¹-1, so any increment overflows.

spec/client_spec.rb
const86: One read carries both responses, so the discard-mid-read case covers separate-read delivery too — nothing from the three blocks is lost.

HoneyryderChuck (approved):

HoneyryderChuck:
thx @const86

